PR #38, opened 2015-02-11
Includes src/<buildType.name>/play as place for resources

saguinav:
After the change included from https://github.com/Triple-T/gradle-play-publisher/pull/37 (which is great :+1: ) I think it makes sense to extend it by including **src/\<buildType.name\>/play** as another place to look for play resources. That would be very convenient for projects like this:

```groovy
buildTypes {
    debug {
       applicationIdSuffix '.debug'
       versionNameSuffix " DEBUG"
    }

    release {
        minifyEnabled true
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
        signingConfig signingConfigs.release
    }

    dogfood.initWith(buildTypes.release)
    dogfood {
        applicationIdSuffix '.dogfood'
        versionNameSuffix " DOGFOOD"
    }
}
```

For publishing a **dogfood** or **release** build version into Google Play it seems more appropriate to include play resources under **src/dogfood/play** or **src/release/play** rather than **src/main/play**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ To use this feature, create a special source folder called ```play```. Inside, c
 ```
 Note: Make sure your texts comply to the requirements of the Play Store, that is they do not exceed the allowed lengths of *30 characters* for the title, *80 characters* for the short description, *4000 characters* for the description and *500 characters* for the summary of recent changes.
 
-Note: You can provide different texts for different locales and product flavors. You may even support additional locales for some product flavors.
+Note: You can provide different texts for different locales, build types and product flavors. You may even support additional locales for some build types or product flavors.
 
 ### Specify the track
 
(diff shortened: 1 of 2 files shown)


mullender:
+1 on this improvement

bhurling:
We have to make sure that files override each other in the same way the android plugin works. Otherwise, nice work!

May I ask: do you publish multiple versions of your APK in the play store?

mullender:
Our workflow is like this:
* every night
 * a dogfood build gets build, a dogfood build has some extra logging and reporting features and its own packagename
 * generate a changelog on the fly
 * publish to play store using this plugin (to a private beta testing channel)
* for a release
 * release build gets build
 * uses hand written changelog checked into repo
 * publish apk to play store alpha channel
 * manually promote to prod
 * update store listing

bhurling:
Fixed in 1.0.2

PR #55, opened 2015-03-24
Defer applicationId retrieval to when the publish task runs

### LINKED ISSUE #54: Deferring applicationId retrieval to when the publishApkTask runs
Right now, the applicationId is captured at configuration and persisted throughout the `PlayPublishTask`'s lifetime, even though a build might dynamically change the applicationId after that. Retrieving the APK file would run into the same issue if the developer set custom APK names, but it doesn't run until the `PlayPublishApkTask` runs, and thus retrieves the correct final name after the app has been assembled. My proposal is to do the same with applicationId. I think it would be an easy change, since the variant instance is held in the `PlayPublishApkTask`. 

In [PlayPublishApkTask][0], instead of using the saved `applicationId` field, we can simply call `variant.applicationId`. I think this should work, though it might have to be `variant.mergedFlavor.applicationId` (I'm not entirely sure if they can ever be different).

If you're open to it, I'd be happy to send a pull request your way.

[0]: https://github.com/Triple-T/gradle-play-publisher/blob/master/src%2Fmain%2Fgroovy%2Fde%2Ftriplet%2Fgradle%2Fplay%2FPlayPublishApkTask.groovy

hzsweers:
This should resolve #54 

This removes the `applicationId` field entirely from `PlayPublishTask` and instead keeps a `variant` field. This way, the applicationId retrieval can wait until the task actually runs, allowing for any user change to the applicationId during project configuration without causing the upload to fail later due to an out-of-date applicationId.

I've added a test as well. Please let me know if I should change/improve/fix it in anyway, as I haven't written a test for groovy before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: android
 
 android:
   components:
-    - build-tools-21.1.2
-    - android-21
+    - build-tools-22.0.1
+    - android-22
 
 script:
   - ./gradlew check --stacktrace
--- a/src/main/groovy/de/triplet/gradle/play/BootstrapTask.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/BootstrapTask.groovy
@@ -30,7 +30,7 @@ class BootstrapTask extends PlayPublishTask {
 
     def bootstrapListing() {
         List<Listing> listings = edits.listings()
-                .list(applicationId, editId)
+                .list(variant.getApplicationId(), editId)
                 .execute()
                 .getListings()
         if (listings == null) {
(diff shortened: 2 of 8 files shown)


bhurling:
I think you have to adjust the `.travis.yml` file to download and install the correct SDK

hzsweers:
Yup, just noticed that and added in 839a686

hzsweers:
That fixed it. Should I squash 839a686 and df96021?

bhurling:
Sure, why not? Keeps the history a little cleaner. Otherwise, nice work!

hzsweers:
squashed them (which, interestingly doesn't clear existing comments when coming from a fork). Anything else?

bhurling:
Closes #54

PR #62, opened 2015-04-21
Fetch version codes for each track during bootstrapPlayResources task

brp-david:
My use-case is the following: I maintain an app with about 70 flavors and all these are not kept at the same version (depends on server-side software among other things). Before I push an update I need to know which version the app currently has so I can pull the relevant "What's new". With this addition I can do this automatically.

This addition saves the version code(s) for each track in a file called version-\<track\> in the play-directory during the bootstrapPlayResources task. For multi-apk apps the version codes are written comma-separated.

Would it be desired to wrap it in a config if the extra files are unwanted?

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/BootstrapTask.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/BootstrapTask.groovy
@@ -4,6 +4,7 @@ import com.google.api.services.androidpublisher.model.Apk
 import com.google.api.services.androidpublisher.model.ApkListing
 import com.google.api.services.androidpublisher.model.Image
 import com.google.api.services.androidpublisher.model.Listing
+import com.google.api.services.androidpublisher.model.Track
 import org.apache.commons.io.FileUtils
 import org.gradle.api.tasks.TaskAction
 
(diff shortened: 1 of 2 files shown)


bhurling:
I don't think that pulling the version code during the bootstrap task is a good idea because of two reasons:

* The bootstrap task is intended to setup the `play` folder with all data required for upload. The version code on the other hand won't ever be uploaded anywhere, because it is implicitly encoded in the APK.
* Running the bootstrap task multiple times overrides the changes you might have made to the listing data and perhaps even checked into version control

So I have to decline this PR, sorry. But perhaps we can think of a way to make the tasks extensible so one can attach custom work to any of the main tasks.

brp-david:
Fair enough! Didn't really consider those consequences, probably mostly because the listing basically never changes for my app (apart from screenshots which aren't downloaded anyway...) 

The part I need "help" with is the Google API, so if we could find a way to expose it relatively easy to a task that could be a way to go and would make a more general solution, I don't think my use-case is very common... I'll ponder the problem!

bhurling:
Declined. But a different solution can be discussed in #67

Issue #64, opened 2015-04-27
Error message on texts which are to long.


Poeschl:
It would be nice to have a error message when you exceed the char limit of the play store fields.
If you like that idea, I will invest some time and do a pull request.

fernandoneto:
:+1:

PR #66, opened 2015-04-29
Error messages

Poeschl:
I implemented some logic to warn the user about the requirements of the play store. #64 
So now you will get an error if the texts exceeds the given limits.
For more look at the Updated README (7d2d02cd0dc2d099506f50476be375e0be396666).

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
@@ -41,15 +41,28 @@ class PlayPublishListingTask extends PlayPublishTask {
             // Check if listing directory exist
             if (listingDir.exists()) {
                 File fileTitle = new File(listingDir, FILE_NAME_FOR_TITLE)
-                def title = TaskHelper.readAndTrimFile(fileTitle, MAX_CHARACTER_LENGTH_FOR_TITLE)
-
                 File fileShortDescription = new File(listingDir, FILE_NAME_FOR_SHORT_DESCRIPTION)
-                def shortDescription = TaskHelper.readAndTrimFile(fileShortDescription, MAX_CHARACTER_LENGTH_FOR_SHORT_DESCRIPTION)
-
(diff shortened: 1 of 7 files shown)


bhurling:
Nice work

bhurling:
Fixes #64

src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
```diff
@@ -41,15 +41,28 @@ class PlayPublishListingTask extends PlayPublishTask {
...
+                    if (!TaskHelper.checkForTextLength(fileTitle, MAX_CHARACTER_LENGTH_FOR_TITLE)) {
+                        throw new LimitExceededException(fileTitle, MAX_CHARACTER_LENGTH_FOR_TITLE)
+                    }
+                    if (!TaskHelper.checkForTextLength(fileShortDescription, MAX_CHARACTER_LENGTH_FOR_SHORT_DESCRIPTION)) {
+                        throw new LimitExceededException(fileShortDescription, MAX_CHARACTER_LENGTH_FOR_SHORT_DESCRIPTION)
+                    }
+                    if (!TaskHelper.checkForTextLength(fileFullDescription, MAX_CHARACTER_LENGTH_FOR_FULL_DESCRIPTION)) {
+                        throw new LimitExceededException(fileFullDescription, MAX_CHARACTER_LENGTH_FOR_FULL_DESCRIPTION)
+                    }
+                    if (!TaskHelper.checkForTextLength(fileVideo, Integer.MAX_VALUE)) {
```
ChristianBecker: Is this check useful?

bhurling:
I think it might be even smarter to move that check into the `TaskHelper.readAndTrimFile()` itself. In fact, there is already a check for that. We just have to pass that boolean with it.

bhurling:
The property itself is very helpful. Without it we would just cut off characters without telling the developer.

ChristianBecker:
I was just curious to check for `Integer.MAX_VALUE` ;) Ok, I'll move this check into the `readAndTrimFile()` method in a follow-up.

PR #68, opened 2015-04-29
Merged checkForTextLength into readAndTrimFile

ChristianBecker:
Follow-up for #66


Poeschl:
This way seems a lot nicer. Thanks for the improvement :+1:

PR #69, opened 2015-04-29
Updated AndroidPublisher dependency from rev16-1.19.1 to rev18-1.20.0

ChristianBecker:
Unfortunately I was unable to find a changelog or release notes, so it's up to you if this gets merged in.


bhurling:
I just diffed the two versions and there is no real difference between rev16 and rev18 aside from some changes to the Javadoc and the updated `google-api-client` dependency. That one just brings some additional checks. So nothing that could break anything.

PR #71, opened 2015-05-01
Use Groovy Beans instead of getters in most places

ChristianBecker:


src/main/groovy/de/triplet/gradle/play/LimitExceededException.groovy
```diff
@@ -14,9 +13,4 @@ class LimitExceededException extends IllegalArgumentException {
     String getMessage() {
         return message;
     }
-
-    @Override
-    String getLocalizedMessage() {
```
ChristianBecker: Defaults to `getMessage()` so there is no need to overwrite this.

PR #80, opened 2015-07-15
Support for staged rollouts

### LINKED ISSUE #79: Support rollout track
When choosing track `rollout` the Publishing API allows specification of a `userFraction` parameter as specified [here](https://developers.google.com/android-publisher/api-ref/edits/tracks) and [here](https://developers.google.com/android-publisher/tracks).

ubuntudroid:
Fixes #79.

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublisherPluginExtension.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublisherPluginExtension.groovy
@@ -24,14 +24,6 @@ class PlayPublisherPluginExtension {
         return track
     }
 
-    private Double userFraction = 0.1
-
-    void setUserFraction(Double userFraction) {
-      this.userFraction = userFraction
-    }
-
-    def getUserFraction() {
-      return userFraction
-    }
+    Double userFraction = 0.1
 
 }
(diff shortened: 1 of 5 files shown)


src/main/groovy/de/triplet/gradle/play/PlayPublisherPluginExtension.groovy
```diff
@@ -24,4 +24,14 @@ class PlayPublisherPluginExtension {
         return track
     }
 
+    private Double userFraction = 0.1
```
ChristianBecker: This could be a Groovy Bean if you remove the `private` modifier and the explicit getter and setter.

bhurling:
Nice work. I will perform a staged rollout for one of my apps at the weekend and merge this one if everything works as expected.

bhurling:
I just released my app as a staged rollout and it works like a charm. One thing though: How would we go about increasing the percentage? That might have to be a new task that does not upload an APK but only edits the userFraction part of a track. Alternatively, we could be smart enough to figure that out if the versionCode has not changed and there is a staged rollout for that version.

ubuntudroid:
Great to hear that it works for you, too! :relieved: 

IMHO *increasing* the percentage is normally done via the developer console as most developers visit it directly before doing that to check that there were no critical bugs or user feedback which would prevent further rollout. Of course one could thing about automating these checks as well - in that case a increase-task would be really nice. However I'm not completely sure about the real demand for such a feature.

ubuntudroid:
How about merging this pull request and then we'll just wait and see if there is demand for an "increase rollout percentage" task?

ubuntudroid:
@bhurling Any ETA for the next release? I'm eager to change to the official version of this plugin instead of using my own custom build one incorporating staged rollout support.

bhurling:
Just released 1.1.3 including this change

ubuntudroid:
Great news! :+1:

PR #95, opened 2015-08-26
updated read me to list all config requirements in a single location

timrosenblatt:
As I was configuring the plugin, I found it helped me out to have a single checklist of all the things I needed to do. I thought it might be helpful to everyone else as well if I PR'd an update to the README. Thanks for making such a useful tool.


bhurling:
Thanks for your contribution. I hope you don't mind me reordering it a bit

Issue #100, opened 2015-10-05
Bootstrapping and Listing tasks should not require signingConfig


Nutomic:
I've set up the plugin as described in the Readme, but all tasks are shown as "not found". They also aren't listed in `./gradlew tasks`. 

Console Output:
```
$ gradle publishApkRelease
Could not find ZipAlign task. Did you specify a signingConfig for the variation Release?

FAILURE: Build failed with an exception.

* What went wrong:
Task 'publishApkRelease' not found in root project 'syncthing-android'.

* Try:
Run gradlew tasks to get a list of available tasks. Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 2.471 secs
```

[Code is here](https://github.com/syncthing/syncthing-android/tree/translation-uploads)

ivanove:
Hi 

if you are on MAC, run this " ./gradlew task" you will find the correct name, because publishApkRelease is just an example, for me the task is "publishMyAppplicationApkRelease" so did like I told you.

Nutomic:
Like I said, I already tried that. `./gradlew tasks | grep publish` is empty.

bhurling:
The plugin only creates those tasks when there is a valid signingConfig for the release type. It even prints a warning if there is something wrong with that signingConfig (see the first output line in your example). 

Please make sure to create a valid keystore and set it up for the release build type.

ivanove:
Look when I try ./gradlew tasks
the gradle output is that :

Play Store tasks
----------------
bootstrapBusinessAppsReleasePlayResources - Downloads the play store
listing for the BusinessAppsRelease build. No download of image resources.
See #18.
bootstrapBusinessReleasePlayResources - Downloads the play store listing
for the BusinessRelease build. No download of image resources. See #18.
generateBusinessAppsReleasePlayResources - Collects play store resources
for the BusinessAppsRelease build
generateBusinessReleasePlayResources - Collects play store resources for
the BusinessRelease build
publishApkBusinessAppsRelease - Uploads the APK for the BusinessAppsRelease
build
publishApkBusinessRelease - Uploads the APK for the BusinessRelease build
publishBusinessAppsRelease - Updates APK and play store listing for the
BusinessAppsRelease build
publishBusinessRelease - Updates APK and play store listing for the
BusinessRelease build
publishListingBusinessAppsRelease - Updates the play store listing for the
BusinessAppsRelease build
[...]

PR #102, opened 2015-10-14
Made signingConfig not require for bootstrapping and lisitng tasks

shiraji:
see #100 

~~I assumed if there are no signingConfig, throw `IllegalArgumentException`. If you don't want to create the publishApk task names, let me know. I will put the validation back to `PlayPublisherPlugin`.~~

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublisherPlugin.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublisherPlugin.groovy
@@ -48,10 +48,6 @@ class PlayPublisherPlugin implements Plugin<Project> {
             def assembleTask = outputData.assemble
 
             def variantData = variant.variantData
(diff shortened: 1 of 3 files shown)


shiraji:
After I took shower, I felt this is not good. I will fix this.

shiraji:
Here you go. I rebased the previous commit. Could you check this pull request, @bhurling ?

bhurling:
This looks very good now. Much better than the first version. Thank you very much.

The only thing we could discuss is the actual condition in the if-clause. Do you still think checking for the zipAlign task is the best solution? Does not affect this PR, though. So I will happily merge this.

shiraji:
Thank you for merging my pull request!

To be honest, while I was developing this feature, I was not sure why this condition required. If there are not good reason to stay the condition, I can remove this condition in different issue.

PR #119, opened 2016-01-15
Play folder configuration via android.SourceSets

### LINKED ISSUE #74: Folder path
The play folder must be in `src/${flavor}/play` but there is any way to configure another path?
Perhaps something like:

```
play {
    track 'production'
    path 'src/customer/${flavor}/play'
}
```

rzhukov:
Closes #74

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublisherPlugin.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublisherPlugin.groovy
@@ -4,6 +4,10 @@ import com.android.build.gradle.AppPlugin
 import org.apache.commons.lang.StringUtils
 import org.gradle.api.Plugin
 import org.gradle.api.Project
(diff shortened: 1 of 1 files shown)


kenyee:
I'd like to see this merged too....seems weird to have stuff in
  src/<flavor>/play
instead of
  src/play/<flavor>

Kazemir:
+1 for this request! I have more then 20 flavors in my app, So I moved all resoursers in dir 'src/flavors' (otherwise it will take a long time to search for a 'main' folder). Now it's looks like 'src/flavors/{flavorName}/res'. So it will be perfect to have 'src/flavors/{flavorName}/play' also.

P.S. It was motivation for speeding up this merge request! 😄

SUPERCILEX:
We've decided against doing this in the plugin, but Gradle will let you [override us](https://github.com/Triple-T/gradle-play-publisher/issues/74#issuecomment-132750435).

PR #122, opened 2016-01-21
Untrack application in alpha when deploying to beta

otnateos:
To allow us to publish to beta when we have app active in alpha,
I remove the tracking of the app from alpha after successful
update on beta channel.

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
@@ -32,6 +32,12 @@ class PlayPublishApkTask extends PlayPublishTask {
         edits.tracks()
                 .update(variant.applicationId, editId, extension.track, newTrack)
                 .execute()
+        if (extension.track?.equals("beta")) {
+          Track emptyTrack = new Track().setVersionCodes([])
+          edits.tracks()
+                  .update(variant.applicationId, editId, 'alpha', emptyTrack)
+                  .execute();
+        }
 
         if (inputFolder.exists()) {
 


bhurling:
I have some concerns about this change:

1. I'd rather not automatically deactivate the APK that sits in the alpha track. If someone accidentally uploads to beta, we break his alpha test.
2. The play web console behaves the same. If you are using the alpha channel and upload a higher version to beta it displays an error. In that case it does not let you apply the changes.
3. I guess we have the same issues between rollout vs. alpha, rollout vs. beta, production vs. all the other tracks

Maybe we should explicitly define tasks that removes APKs from tracks. Then a user can decide whether he wants his publishApk task to depend on one or more of those untrack tasks (or if he'd rather keep the old behaviour as a security measure).

otnateos:
> I'd rather not automatically deactivate the APK that sits in the alpha track. If someone accidentally uploads to beta, we break his alpha test.

That sounds fair, some people may accidentally do that

> The play web console behaves the same. If you are using the alpha channel and upload a higher version to beta it displays an error. In that case it does not let you apply the changes.

That's quite different from my experience. Normally when I upload a higher version to beta, Play web console deactivate my alpha automatically. I believe that is outlined in https://support.google.com/googleplay/android-developer/answer/3131213?hl=en#availability under point 2

> I guess we have the same issues between rollout vs. alpha, rollout vs. beta, production vs. all the other tracks

Yes, most likely. I was looking for a quick fix and like in the issue ticket I mentioned  https://developers.google.com/android-publisher/tracks#staged_rollouts , about to send empty version code to rollout once it is ready for production. That's why i thought it should happen automatically

> Maybe we should explicitly define tasks that removes APKs from tracks. Then a user can decide whether he wants his publishApk task to depend on one or more of those untrack tasks (or if he'd rather keep the old behaviour as a security measure).

Sounds like a good solution

bhurling:
Closing because we are going to add separate tasks to untrack APKs

PR #123, opened 2016-01-26
allow to provide p12 password

broglep-koubachi:
This pull request allows to provide a custom p12 password, either with

```
play {
    serviceAccountEmail 'service'
    pk12File file('file.p12')
    pk12Password 'asecret'
}
```

(using password for both store & key, and using default `privatekey` alias)

or more customised

```
play {
    serviceAccountEmail 'service'
    pk12File file('file.p12')
    storePassword 'storepass'
    keyAlias 'keyalias'
    keyPassword 'keypass'
}
```

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/AndroidPublisherHelper.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/AndroidPublisherHelper.groovy
@@ -20,7 +20,8 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory
+import com.google.api.client.util.SecurityUtils;
 import com.google.api.services.androidpublisher.AndroidPublisher;
 import com.google.api.services.androidpublisher.AndroidPublisherScopes;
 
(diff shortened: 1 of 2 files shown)


rocboronat:
+1!

bhurling:
It looks like you forgot to commit your changes to the `PlayPublisherPluginExtension`

bhurling:
I am sorry, but I don't fully understand that change. Your version now calls `setServiceAccountPrivateKey()` as well as `setServiceAccountPrivateKeyFromP12File()` on the same `GoogleCredential.Builder()` instance. Which means that the latter will always override what was set by the former.

In addition, I don't know how and why you would create a p12 that is different from the one you can download via the Google API Console. Can you explain your Use Case?

broglep-koubachi:
you are right, fixed that now. 

Well we create a different p12 file for security purposes. Due to the permission model in Google Play, the service account needs quite broad permissions to work, and having the key lying around essentially unencrypted (the default password is well known), is not something that needs to be prevented, thus we re-encrypt the private key with a proper secret

giovannicandido:
Really useful for continuous integration build. I'm would like very much to use this feature.
For security the unsecured key file should not be committed, is best practice to use secured env variables with passwords.

lolgesten:
This is really needed for things like deploying from CircleCI where we don't want an unprotected .p12 with full store access being checked into the repo.

luongvo:
👍

SUPERCILEX:
Thanks for the PR! Unfortunately, we've decided against going that route and will instead rely on whatever built-in mechanisms various CI servers have to encrypt data. See #271 for progress.

PS: in the future, please make sure to get a green light from an issue before making a PR so your work doesn't have to be abandoned. 😊

PR #124, opened 2016-01-31
Adding a clarification of the userFraction param in the README.md

rocboronat:
Following the readme, I've seen that the default userFraction was 0.1. I thought it was a porcentage, like in the Google Play environment. But 1 meant 100%, not 1%... so I propose to add this small clarification, to avoid doing the mistake of publishing a version at 50% instead of 0.5%, as I did :·)


bhurling:
Thanks, even the official API reference is not very clear about that parameter.

Issue #125, opened 2016-01-31
Don't break if lowering the userFraction


rocboronat:
Usually, we launch a version to 0.05% of the target and check them daily using Crashlytics. And we increase that porcentage daily. Usually, we're at 10%, and we decide to launch a hotfix to Google Play.

In that case, the publisher breaks, 'cause we're trying to lower the userFraction.

It's possible to only use this value when the rollout process is new? I mean, if we're at 10%, it's ok that the .APK goes to 10% instead of 0.05% of the target.

EsteveAguilera:
+1

SUPERCILEX:
This isn't something the plugin controls. However, v2.0 will support halting a release which is all the Play Store allows. Addressed in #291.

PR #110, opened 2015-10-22
Added changeUserFraction that change userFraction

shiraji:
see #104 

This one may conflict with #108

PR #108, opened 2015-10-19
Added promoting track tasks

shiraji:
see #103 

Feel free to comment on this pull request. I will remove [WIP] when this pull request is ready.

These commit will be rebase anytime I think it's necessary.

Issue #104, opened 2015-10-15
Change userFraction after publishing apk


shiraji:
`gradle-play-publisher` has `userFraction` but it only useful when user publishs apk. I want to change userFraction after publishing apk.

`./gradlew changeUserFraction -PuserFraction=0.5`

This may be a different task but I want to pass userFraction as a parameter. Because I don't want users to change `build.gradle` every time executing this command. 

Creating tasks for all userFraction like `changeUserFractionTo05` could be possible work around.

SUPERCILEX:
This plugin is meant to support continuous integration, not be a replacement for the Play Console. I think updating the user fraction could easily be done in the Google Play Console Android app or on the website. @bhurling what do you think?

shiraji:
I really didn't know this plugin is only for supporting CI. You need to close #108 and #110, too. I created this issue and PR because I run this plugin in command line. I wanted to keep changing fraction based on the feature I would like to release.

SUPERCILEX:
I wouldn't say it's _only_ meant CI, but we're spread a little thin and trying to mirror the Play Console would be major overkill IMO. However, @bhurling has the final say so I'd wait for his opinion. 😄

bhurling:
I agree that batch updating the user fractions can be very useful for maintainers of whitelabel Apps. And that's basically why this plugin was created in the first place. Not sure about the API, though. @gtcompscientist do you have any thoughts?

gtcompscientist:
@bhurling I think that there are a lot of updating tasks that don't include the APK would be useful for people managing projected with lots of variants.

PR #133, opened 2016-03-11
Added a class that returns the next rollout user fraction

rocboronat:
In the way to fix the issue #125 , I started creating a class that gives you the next user fraction at Google Play. So, if you say "hey, I'm at 0.05%", it says "ho, the next rollout fraction is "0.1%".

There are my first groovy lines. Feel free to change whatever you need to meet any naming convention you like. By the way, the lines I'm adding to the project are fully tested. No one are calling them yet, but they will be useful for sure.

If you want me to improve something, don't hesitate to share it! :·)

Thanks for your work.

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayRolloutFractionsExtension.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayRolloutFractionsExtension.groovy
@@ -2,24 +2,14 @@ package de.triplet.gradle.play
 
 class PlayRolloutFractionsExtension {
 
-    def fractions = [0.005, 0.01, 0.05, 0.1, 0.2, 0.5, 1.0]
+    def fractions = [0.005d, 0.01d, 0.05d, 0.1d, 0.2d, 0.5d, 1.0d]
 
     double getNextUserFraction(double userFraction) {
-        switch (userFraction) {
-            case 0.005:
-                return 0.01;
-            case 0.01:
-                return 0.05;
-            case 0.05:
-                return 0.1;
-            case 0.1:
-                return 0.2;
-            case 0.2:
-                return 0.5;
-            case 0.5:
-                return 1.0;
-            default:
-                return 1.0;
+        if (userFraction == 1.0d) {
+            return 1.0d;
+        } else {
+            int userFractionIndex = fractions.indexOf(userFraction);
+            return fractions.get(userFractionIndex + 1);
         }
     }
 
--- a/src/main/groovy/de/triplet/gradle/play/PlayRolloutFractionsExtension.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayRolloutFractionsExtension.groovy
@@ -5,7 +5,22 @@ class PlayRolloutFractionsExtension {
     def fractions = [0.005, 0.01, 0.05, 0.1, 0.2, 0.5, 1.0]
 
     double getNextUserFraction(double userFraction) {
-        return 0.01
+        switch (userFraction) {
+            case 0.005:
+                return 0.01;
+            case 0.01:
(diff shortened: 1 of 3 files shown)


src/main/groovy/de/triplet/gradle/play/PlayRolloutFractionsExtension.groovy
```diff
@@ -0,0 +1,18 @@
+package de.triplet.gradle.play
+
+class PlayRolloutFractionsExtension {
+
+    def fullRelease = 1.0d;
+
+    def fractions = [0.005d, 0.01d, 0.05d, 0.1d, 0.2d, 0.5d, fullRelease]
```
ChristianBecker: We should be consistent about the trailing semicolon

ChristianBecker:
What will happen if you call `getNextUserFraction(0.02d)`? I guess it will return `0.005` in that case. I'm not sure if this is expected behavior.

rocboronat:
Thanks for the feedback, @ChristianBecker !

I just removed all the trailing semicolons, and added a test that does your check and tests that the new method launches an IllegalRolloutException when it recieves an invalid rollout value. I think that it's the best thing that we can do to clarify why we're crashing.

rocboronat:
Any feedback about this PR?

rocboronat:
@bhurling please, consider this PR. It's fully tested and it will be useful for the purpose of the project for sure. In addition, it loves your code standards :·D

bhurling:
This class is nice but apparently it's not used in any task yet. I could think of a task or script that promotes an APK in a staged rollout to the next level. I just wonder if this gradle plugin should really handle these things. Promoting APKs isn't something that is part of the regular build cycle.

cc @ChristianBecker

rocboronat:
Yes @bhurling , this class is not used... 'cos it's new in this PR :·)

By the way, this plugin is wide enough to publish APK's to Google Play, download metadata from Google Play, upload pictures to the app page... so, adding another feature related with publishing to Google Play is good idea.

By the way, how to use the plugin is up to you and your teammates. In our case, we at http://quitnowapp.com upload at 0,5% and go to next level every day. It may be great if the plugin helps us automatizing it, leaving the humans just the task to follow the progressive upload using Fabric.

ChristianBecker:
I'm not sure if we should get this PR merged as-is with code that isn't used in any task yet, as @bhurling already mentioned.

One way to integrate it in a meaningful way could be that we merge this PR with #110 so that we have a task that automatically increases the current `userFraction` if we are on the `promote` track.

This could play nicely with #108 so that we have new `promoteTrack${variationName}` and 
`increaseUserFraction${variationName}` tasks.

Any thoughts, @bhurling, @rocboronat, @shiraji ?

rocboronat:
If it doesn't help anyone, it's better to avoid merging them. By the way, it's easy enough to be retyped at some point of the future... Actually, it was just my hello world in Groovy, and I prefer to share something with the world than just doing a dummy app to try a language :·)

Going to the topic again, totally agree to join this PR with #108 or #110. It will be nice if the task to pass to the next rollout fraction is just `increaseUserFraction`, skipping to say "going to 5% or 10%".

By the way: happy to see that this feature is coming back to topic! 🎉

bhurling:
@rocboronat @ChristianBecker I agree that this feature adds quite some value to the plugin. This should go into one of the next releases. And indeed it should be a combination of all the track/rollout related proposals.

One thing to keep in mind though, is that we now support ABI splits. That means we might have a group of APKs that should be promoted at the same time.

SUPERCILEX:
Hi there, we've moved to Kotlin so this PR is unfortunately obsolete. Sorry about that! 😊 See #104 for discussion.

PR #134, opened 2016-03-11
Testing ImageFileFilter with some tests, to avoid breaking it in the future

rocboronat:
In order to do an useful kata-like exercise, I created some tests for ImageFileFilter. Hope you find them  useful, too!


bhurling:
Thanks a lot. More tests are always better

PR #136, opened 2016-04-07
Fix character counting

cypressious:
Special characters like umlauts are counted as more than one character while Google counts them as only one. Use BreakIterator for better counting.


ChristianBecker:
The build is failing. Can you fix this please?

Can you also add a test that would fail without this fix?

cypressious:
Broke another test, fixing again...

cypressious:
The test that is failing on Travis is passing on my machine. I think it has something to do with git's character encoding.

cypressious:
Got it, it was actually the line breaks, not the umlauts.

cypressious:
@ChristianBecker Would you like to take another look?

bhurling:
LGTM, I will just restructure a bit

PR #137, opened 2016-04-18
Add ability to specify play data directory

kurtacious:
This pull request adds the ability to specify the directory the play store data is stored.  This is done by assigning a directory path to `metadataRoot` in the `play` block.

For example:

```
play {
   ...
   metadataRoot = file("../play_metadata")
}
```

In our case, we have multiple configurations that are generated at build time (not flavors).  This will allow us to maintain a directory for each configuration which is chosen at build time using a gradle system property (e.g. `./gradlew publishRelease -Dconfig_id=<config_id>`).  In our case we'd set metadataRoot to something like `file("../custom_build_configs/${config_id}/play_metadata")`.


kurtacious:
I believe this is an alternative solution to what #119 provides.

PR #142, opened 2016-04-28
Fix regex for Filipino locale for publishing

enterline:
- Noticed that when I run a `publishListing`, it was not updating the `fil` locale with screenshots.
- Fixed the matcher regex to include the `fil` locale.
- I am not an expert in regex, but this seems to work.
- Added a couple of tests to verify


bhurling:
Thanks. Didn't know about that one

Issue #76, opened 2015-06-02
Multiple Service Accounts


robertoestivill:
I'm working on an application that has two flavours. Each flavour corresponds to a brand/client. Each brand has it's own developer account.

Is it possible to configure a per-flavour service account? do you have plans to add this functionality? 

Do you guys know how can I accomplish this in the current state? 
Maybe with a task that configures the play plugin for each flavour before actually running the upload?

Thanks

robertoestivill:
Ended up using jvm variables similar to the following workaround in #65 
Would still be nice to know what the collaborators opinion is on having this per signing config.

bhurling:
I thought of something similar to Android's signingConfigs. So one can define one or more `playSigninigConfigs`and then use it inside of the build flavour block. I think we would have to extend their DSL to do something like that. Is that even possible?

mattxander12:
I've sort of worked around this issue myself in a different way. I too have one project with many flavors that go to different Google Play accounts. 

What I've done, is in my build.gradle file, specify signingConfigs for each dev account (even if the keystore info is the same) for each flavor config. 

I created some property files based on the signingConfig name with the dev service account info in it, and dropped them into the module root directory. 

Then in 'PlayPublishApkTask.groovy', I added the following to the publishApk() function: 
```groovy
 @TaskAction
    publishApk() {
        def signingConfigName = variant.signingConfig.name;
        def propertiesFileName = "${signingConfigName}.play.properties"
        
        File propFile = new File(propertiesFileName.toString());
        
        if (propFile.exists()) {
            Properties properties = new Properties()
            properties.load(new FileInputStream(propFile))

            extension.serviceAccountEmail = properties['serviceAccountEmail']
            extension.pk12File = new File(properties['pk12File'])
        }

        super.publish()
        ... 
}
```

This gives the result of the correct dev service account info feeding in, if available, else it'll use the standard 'play' closure configuration in build.gradle. 

Not worrying about having to switch service account information when submitting app updates was my main goal. It's a bit messy and requires some extra setup, but working with limited Android DSL property availability, this was an alternative solution I came to.

robertoestivill:
Hi @mattxander12 . Your solution sounds like a better option. Do you mind sharing a repo with all the needed directories/files ? Thanks

rubeus90:
Hello,

[...]

PR #146, opened 2016-05-28
Add Untrack APK task

otnateos:
This commit adds untrackApk task for the plugin to allow user to untrack
app from another Play Store channel when they are using 2 different
channels to deploy their app.

```
gradle untrackApkRelease
```

This commit adds `untrack` and `untrackFormat` properties to the
extension to allow further customisation on which version of the app to
be disabled from Play Store.

```groovy
play {
  // untrack from alpha any active versions between 300 and 399
  untrack = 'alpha'
  untrackFormat = '^3[0-9]{2}\$'
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ The plugin creates the following tasks for you:
 * `publishApkRelease` - Uploads the APK and the summary of recent changes.
 * `publishListingRelease` - Uploads the descriptions and images for the Play Store listing.
 * `publishRelease` - Uploads everything.
+* `untrackApkRelease` - Untrack APK from Play Store
 * `bootstrapReleasePlayResources` - Fetches all existing data from the Play Store to bootstrap the required files and folders.
 
 Make sure to set a valid `signingConfig` for the release build type. Otherwise, there won't be a publishable APK and the above tasks won't be available.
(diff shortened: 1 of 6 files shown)


danmattsson:
Sorely needed to enable automated builds - please consider pulling this.

bhurling:
@otnateos Thank you for that contribution. Do you think it is a good idea to put `untrack` and `untrackFormat` into the `play` closure? That way you'd have make changes to your build.gradle anytime you want to untrack from another channel. But I like the idea with the regex

danmattsson:
Valid point, @bhurling. I don't know if our approach would address your issue but we configure play this way:

`    productFlavors {
        flavorDimensions "environment"

        production {
            dimension "environment"
            googlePlayTrack = 'production'
            googlePlayUntrack = 'beta'
        }
        beta {
            dimension "environment"
            googlePlayTrack = 'beta'
            googlePlayUntrack = 'alpha'
        }
        alpha {
            dimension "environment"
            googlePlayTrack = 'alpha'
        }
        dev {
            dimension "environment"
        }
    }

    play {
        track = googlePlayTrack
        untrack = googlePlayUntrack
        jsonFile = file('gplay-api-access-keys.json')
        uploadImages = true
    }
`

bhurling:
That's very smart. And it could even help with #76 

I don't see any untrackFormat in your build.gradle. That would be the way to basically skip the untrack task, right? I will put that into the README

bhurling:
I also see another small issue with that solution. You can only untrack one channel at a time. In case you want to push to production I guess you'd have to both untrack alpha and beta.

So let me propose another approach:

1. in the play config block there would be a boolean `untrackOld`
2. if true the untrack task untracks everything that would be blocking the new APK (anything with a lower version code in any of the lower tracks)

Example: pushing version 42 to production would untrack every APK with a version code less than or equal to 41 in alpha and beta

danmattsson:
That would for sure solve _my_ problem and I guess most others' as well. The simplicity of your proposal far outweighs the potential benefits of being able to filter versions to "untrack".

otnateos:
> I also see another small issue with that solution. You can only untrack one channel at a time. In case you want to push to production I guess you'd have to untrack both alpha and beta.
> 
> So let me propose another approach:
> 
> in the play config block there would be a boolean untrackOld
> if true the untrack task untracks everything that would be blocking the new APK (anything with a lower version code in any of the lower tracks)

@bhurling Yes, it will make it simpler. Which I did in PR #122 minus the flag 

> if false we keep the old behaviour which stops the process if there is some blocking APK
> Example: pushing version 42 to production would untrack every APK with a version code less than or equal to 41 in alpha and beta

However the reason I didn't go that path and opted for regex approach is due to backward version compatibility. If you guys notice from API 23 you can't upload newer version with lower API support. From my previous project, we had to keep the old API support while having new one as well.

E.g. Version 150 is used for up to API 21 , while Version Code 1011 used for API 23

PR #150, opened 2016-06-08
Update PlayPublishListingTask.groovy

ctsdevmumbai:
Print statements


src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
```diff
@@ -47,6 +47,8 @@ class PlayPublishListingTask extends PlayPublishTask {
             def locale = dir.name
 
             File listingDir = new File(dir, LISTING_PATH)
+            System.out.println("Listing Path: " + LISTING_PATH + " , Exixts: " + listingDir.exists())
```
enterline: Should be `Exists`

ChristianBecker:
Can you explain the use case / motivation for this PR?

PR #160, opened 2016-08-10
Add support for uploading obb files

paour:
Fix the project name (hard-coded rather than derived from the top-level directory name)


src/main/groovy/de/triplet/gradle/play/PlayPublishObbTask.groovy
```diff
@@ -0,0 +1,47 @@
...
+
+class PlayPublishObbTask extends PlayPublishTask {
+
+    File inputFolder
+
+    @TaskAction
+    publishObb() {
+        super.publish()
+
+        if (!extension.uploadObbMain && extension.uploadObbPatch) {
```
ChristianBecker: I think this had to be `… && !extension.uploadObbPatch` ?

paour:
@ChristianBecker do you intend to merge this pull request at some point (should I resolve the conflicts)?

paour:
No response in 20 days, we can't wait and have to fork, which is really too bad…

PR #161, opened 2016-08-12
Working DSL for multiple service accounts

brp-david:
Handling multiple service accounts per #76 
Got the DSL working also added possibility of adding a default playAccountConfig to the defaultConfig-block (see PlayPublisherPluginTest.testPlaySigningConfigs) 

Not sure about the handling of the config in the plugin.apply, could probably be handled better. But couldn't figure out a better way to access it in the publish-task.


brp-david:
Any input or something blocking this? One thing I've thought about is that the old behavior doesn't work so config would have to be changed when updating. Making it possible to fallback to the old way would be possible of course.

brp-david:
Added backwards compatibility so config doesn't need to change. I have a few tests but I don't really feel like they test the "real" thing. I'm guessing it would need some mocked test to check which config is actually used by the publish task but I'm not sure how to go about that.

brp-david:
I guess that pretty much proved my point, had missed a row that would have made this not work and tests did not pick it up. So missing is a test that actually tests the used config. I have tested manually with my own Android-project and verified both the new style and old style config.

clhols:
I like these DSL changes. I have cherry-picked it and used it for a month. Very nice!

src/test/groovy/de/triplet/gradle/play/PlayPublisherPluginTest.groovy
```diff
@@ -133,6 +119,106 @@ class PlayPublisherPluginTest {
...
+                    playAccountConfig = playAccountConfigs.free
+                }
+                paid {
+                    playAccountConfig = playAccountConfigs.paid
+                }
+            }
+        }
+        project.evaluate()
+
+        assert project.tasks.bootstrapDefaultFlavorReleasePlayResources.playAccountConfig.serviceAccountEmail == 'default@exmaple.com'
```
ChristianBecker: For the unit test we normally use the static methods from `org.junit.Assert` rather then the `assert` keyword.

src/test/groovy/de/triplet/gradle/play/PlayPublisherPluginTest.groovy
```diff
@@ -133,6 +119,106 @@ class PlayPublisherPluginTest {
         }
     }
 
+    @Test
+    public void testJsonFileBackwardsCompatibility() {
+        Project project = TestHelper.evaluatableProject()
```
ChristianBecker: Those types can be inferred by using the `def` keyword

ChristianBecker:
@brp-david Thanks for your fast response. I hope I can test your changes soon, so that we can get this PR merged!

src/main/groovy/de/triplet/gradle/play/PlayPublisherPlugin.groovy
```diff
@@ -59,6 +69,11 @@ class PlayPublisherPlugin implements Plugin<Project> {
             // Create and configure publisher meta task for this variant
             def publishListingTask = project.tasks.create(publishListingTaskName, PlayPublishListingTask)
             publishListingTask.extension = extension
+            if (!variant.flavorName.isEmpty()) {
+                publishListingTask.playAccountConfig = project.android.productFlavors[variant.flavorName].playAccountConfig
+            } else {
+                bootstrapTask.playAccountConfig = project.android.defaultConfig.ext.playAccountConfig
```
bhurling: this should be `publishListingTask.playAccountConfig` I guess

bhurling:
Man, this is amazing. Just ran a few tests and it works like a charm. Great job.

bhurling:
Small issue is fixed and I added two assertions in the test.

Available in `1.2.0-beta3`

brp-david:
Excellect, thanks!

PR #167, opened 2016-09-16
Feature/validation

Malinskiy:
I noticed that there is no way to validate text without trying to upload the listing.
E.g. you're doing checks on CI and you need to validate and reject translations if they're not valid.

I didn't have time to write tests


SUPERCILEX:
Hi there, we've moved to Kotlin so this PR is unfortunately obsolete. Sorry about that! 😊 Please open an issue so we can discuss this further and decide how much metadata validation we want on the plugin's side of things

PR #169, opened 2016-09-20
Untracking throws exception

jaynakus:
Code should omit exception thrown when there is no active version in track.

`Execution failed for task ':App:publishApkAgentRelease'.

> com.google.api.client.googleapis.json.GoogleJsonResponseException: 404 Not Found
>   {
>     "code" : 404,
>     "errors" : [ {
>       "domain" : "androidpublisher",
>       "message" : "Track not found: no APK version codes in the specified track.",
>       "reason" : "trackEmpty"
>     } ],
>     "message" : "Track not found: no APK version codes in the specified track."
>   }
> `

### DIFF
--- a/src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
+++ b/src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
@@ -1,6 +1,7 @@
 package de.triplet.gradle.play
 
 import com.android.build.gradle.api.ApkVariantOutput
+import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.FileContent
 import com.google.api.services.androidpublisher.model.Apk
 import com.google.api.services.androidpublisher.model.ApkListing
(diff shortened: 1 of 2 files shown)


src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
```diff
@@ -45,12 +46,19 @@ class PlayPublishApkTask extends PlayPublishTask {
         if (extension.untrackOld && !"alpha".equals(extension.track)) {
             def untrackChannels = "beta".equals(extension.track) ? ["alpha"] : ["alpha", "beta"]
             untrackChannels.each { channel ->
-                Track track = edits.tracks().get(variant.applicationId, editId, channel).execute()
-                track.setVersionCodes(track.getVersionCodes().findAll {
-                    it > apk.getVersionCode()
-                });
-
-                edits.tracks().update(variant.applicationId, editId, channel, track).execute()
```
bhurling: If possible I'd rather check whether `versionCodes` is empty instead of running into an exception that could have been avoided.

bhurling: I see, it's the get request that's failing. I tested and the changes work fine. Merging

PR #182, opened 2016-12-27
Support for universal apk when splits API is enabled.

PiDyGB:
The universal apk is skipped and not uploaded to Google Play Store.

### DIFF
--- a/src/test/groovy/de/triplet/gradle/play/PlayPublishTaskTest.groovy
+++ b/src/test/groovy/de/triplet/gradle/play/PlayPublishTaskTest.groovy
@@ -105,6 +105,38 @@ public class PlayPublishTaskTest {
     }
 
     @Test
+    public void testAbiSplits() {
+        Project project = TestHelper.evaluatableProject()
+        project.play {
+            track 'production'
+            untrackOld true
+        }
+
+        project.android {
+            splits {
+                abi {
+                    enable true
+                    reset()
+                    include 'armeabi-v7a', 'arm64-v8a', 'mips', 'x86', 'x86_64'
+                    universalApk true
+                }
+            }
+        }
+
+        project.evaluate()
+
+        // Attach the mock
+        project.tasks.publishApkRelease.service = publisherMock
+
+        // finally run the task we want to check
+        project.tasks.publishApkRelease.publishApks()
+
+        // verify that we init the connection with the correct application id
+        verify(editsMock).insert("com.example.publisher", null)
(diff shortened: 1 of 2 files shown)


src/test/groovy/de/triplet/gradle/play/PlayPublishTaskTest.groovy
```diff
@@ -105,6 +105,38 @@ public class PlayPublishTaskTest {
...
+
+        // Attach the mock
+        project.tasks.publishApkRelease.service = publisherMock
+
+        // finally run the task we want to check
+        project.tasks.publishApkRelease.publishApks()
+
+        // verify that we init the connection with the correct application id
+        verify(editsMock).insert("com.example.publisher", null)
+        verify(apksMock, times(5)).upload(eq('com.example.publisher'), eq('424242'), any(FileContent.class))
```
bhurling: Shouldn't this upload 6 APKs? 5 ABIs plus one universal?

bhurling:
This will skip the universal APK even though it exists, correct? I don't understand the use case here. I think we should in fact upload the universal APK if it is created. If users don't need that APK they can just say `universalApk false` in the gradle file.

ireneduke:
@bhurling This would be really useful for me. It would be great if this plugin could provide the option for uploading or not uploading the universal apk.

I build the universal apk and upload it to HockeyApp to provide a single release candidate to our internal QA team, but uploading it to Google Play is problematic. The native libraries I use don't include the same set of architectures. For example, one library includes mips and the other doesn't. If someone with a mips device downloads the universal apk from Google Play, it will crash because both native libraries don't include it.

bhurling:
I see. So in fact, we'd need a way to tell the plugin whether or not to publish the universal APK. With this changeset alone we'd suddenly stop publishing universal APKs which might be dangerous. So maybe we should introduce an attribute like `skipUniversalApk` and take that into account when we find the APK outputs to publish.

gtcompscientist:
@PiDyGB and @ianduke I would suggest adding a command-line flag to builds to allow you to configure when the `universalApk` is built similar to this:

At the top of your app-level `build.gradle`:

```
def includeUniversal = false
if (project.hasProperty('includeUniversal')) {
    includeUniversal = project.properties['includeUniversal'].toString() == "true"
}
```

Then in the `splits` section:

```
    splits {
        abi {
            ...
            universalApk includeUniversal
        }
    }
```

At the command-line you would use it like this:
`./gradlew [compile/assemble/publish/etc task] -PincludeUniversal=(true,false)`

Issue #163, opened 2016-08-22
Publish APK without listing - FileNotFoundException


Dean79000:
I publish APK only (without listing)

```
apply plugin: 'com.github.triplet.play'
play {
    serviceAccountEmail = 'myemail'
    pk12File = file("path-to-mycert")
    track = 'production'
}
```

When i run task `"publish<Flavor>Release"`, the build fail

```
:app:publish<Flavor>Release
:app:publishListing<Flavor>Release FAILED  
java.io.FileNotFoundException: [...]app/build/outputs/play/<Flavor>Release
```

But if i create empty folder "src/main/play" build is successful. I expect to have successful build without create this folder.

Tested in version 1.1.5

bhurling:
You could only run publishApk<Flavor>Release to work around that issue.
That way the listing task won't be executed. But you have a point there. An
non-existing play folder should be treated the same as an empty one.

damianpetla:
I also have this error but I checked that APK was successfully uploaded to GooglePlayConsole

RyotaMurohoshi:
I failed publishListingRelease task with FileNotFoundException in CI environment.

I think this issue may be related to this issue.

https://github.com/Triple-T/gradle-play-publisher/issues/180

clhols:
Created pull request here:
https://github.com/Triple-T/gradle-play-publisher/pull/184

PR #184, opened 2017-01-16
Issue 163 - Publish APK without listing - FileNotFoundException

clhols:
Fix for issue 163:
https://github.com/Triple-T/gradle-play-publisher/issues/163


src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
```diff
@@ -41,7 +41,7 @@ class PlayPublishListingTask extends PlayPublishTask {
         super.publish()
 
         // Matches if locale have the correct naming e.g. en-US for play store
-        inputFolder.eachDirMatch(matcher) { dir ->
+        if (inputFolder.exists()) inputFolder.eachDirMatch(matcher) { dir ->
```
ChristianBecker: In this case, the following statements from line 131 would be executed. What do you think about "inverting the flow" by
```
if (!inputFolder.exists()) {
    return
}
```

clhols: Yes, it is better to avoid executing any code when the inputFolder doesn't exist.

src/main/groovy/de/triplet/gradle/play/PlayPublishListingTask.groovy
```diff
@@ -40,6 +40,10 @@ class PlayPublishListingTask extends PlayPublishTask {
     publishListing() {
         super.publish()
 
+        if (!inputFolder.exists()) {
```
ChristianBecker: We can move this code above the `super.publish()` call to avoid executing even more code

ChristianBecker:
Thanks!

PR #192, opened 2017-03-10
Update Gradle wrapper to 3.4.1 and Android Gradle plugin to 2.3.0

ChristianBecker:
I think we can take this step and move to the Android Gradle plugin 2.x.y

/cc @bhurling


.travis.yml
```diff
@@ -1,19 +1,23 @@
 language: android
 
-# https://github.com/travis-ci/travis-ci/issues/6040#issuecomment-219367943
 android:
   components:
     - tools
     - platform-tools
-    - tools
```
CristianGM: Do you think is safe to remove the duplicated "tools", today it works because TravisCI is up-to-date but I won't remove it.

ChristianBecker: You are probably right. I'll revert those lines!

Issue #193, opened 2017-03-15
Error:ProductFlavor with name 'XXX' not found.


gdemiollis:
Hi,

Since beta3, I have this Error :

```Error:ProductFlavor with name 'remoteProd' not found.```

In my project, I use different flavors dimension

```groovy
flavorDimensions "mock", "server"

    productFlavors {
        mock {
            dimension "mock"
        }
        remote {
            dimension "mock"
        }
        local {
            dimension "server"
        }
        preprod {
            dimension "server"
        }
        prod {
            dimension "server"
        }
    }
```

My variants are
   - remoteProdRelease
   - remoteProdDebug
   - remotePreprodRelease
   - remotePreprodDebug
   - mockLocalDebug

I don't have this error in beta2

bhurling:
Looks like it's related to having multiple flavor dimensions. I'll write some tests to see what's going on

bhurling:
@brp-david looks like `project.android.productFlavors[variant.flavorName].playAccountConfig` can't handle multiple flavor dimensions. Any idea how to fix that?

brp-david:
I'll look into it, but a quick guess the playAccountConfig has to be stored on the composite flavor name somehow.

bhurling:
I added a failing test: https://github.com/Triple-T/gradle-play-publisher/tree/fix-service-accounts-for-dimensions

bhurling:
@gdemiollis If you don't need multiple service accounts you might want to fall back to beta2 in the meantime

gdemiollis:
Thank you, that's what I will do.

ChristianBecker:
This is fixed in the latest `master` and will be released with the next `beta` release

PR #194, opened 2017-03-15
Fixed playaccountconfig for dimensions

brp-david:
#193 
Solved the dimension problem and should be working fine for dimensions, picking the first config found from the flavors in the variant as per how dimension-overriding should work. I have no idea how to actually use dimensions practically though so I'm unsure how to write a test that actually tests it.


ChristianBecker:
@brp-david Yeah, sorry, missed it! Thanks!

PR #200, opened 2017-03-22
More descriptive message when text exceeds length

### LINKED ISSUE #172: More descriptive message when text exceeds length
Pretty small thing, but useful:

When I try to publish and one of my `whatsnew` files are too long, I cannot see clearly from message which language (or languages) is causing this.
Ex:

```
* What went wrong:
Execution failed for task ':App:publishApkRelease'.
> File 'release -> whatsnew' has reached the limit of 500 characters.
```

I tried to run with `--info` and couldn't find it there. Haven't tried `--debug` as even info was pretty spamy and I'd personally like to see this information by default if possible.

ChristianBecker:
Include the complete relative path of the file that exceeds its limit in the IllegalArgumentException. Fixes #172

FYI @bhurling


bhurling:
Can we think of a better way to make the logger available here? Maybe having a static helper class is not a good idea anyway.

PR #202, opened 2017-03-23
Migrate sources from Groovy to Kotlin

ChristianBecker:


build.gradle
```diff
@@ -51,3 +52,11 @@ afterEvaluate { project ->
         }
     }
 }
+
+// Compile Kotlin then Groovy
```
ChristianBecker: This will be removed once the migration is completed.

src/main/kotlin/de/triplet/gradle/play/PlayPublishListingTask.kt
```diff
@@ -0,0 +1,166 @@
...
+
+                // Just rethrow everything else.
+                throw e
+            }
+
+            // By default this will be skipped – can be enabled in the play extension
+            if (extension.uploadImages) {
+                // Only one ContentFile allowed for featureGraphic
+                val featureGraphicContent = getImages(listingDir, "$IMAGE_TYPE_FEATURE_GRAPHIC/")
+                uploadImages(featureGraphicContent, locale, IMAGE_TYPE_FEATURE_GRAPHIC, 1)
```
ChristianBecker: This is a behavior change: We log a warning if more then one files are found and only a single graphic is expected.

src/main/kotlin/de/triplet/gradle/play/AndroidPublisherHelper.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import com.google.api.client.auth.oauth2.Credential
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
+import com.google.api.services.androidpublisher.AndroidPublisher
+import com.google.api.services.androidpublisher.AndroidPublisherScopes
+import java.io.File
+
+class AndroidPublisherHelper {
+    companion object {
+        fun init(extension: PlayPublisherPluginExtension, config: PlayAccountConfig?): AndroidPublisher {
```
ChristianBecker: We can move this static method into `PlayPublishTask`…?

src/main/kotlin/de/triplet/gradle/play/PlayPublisherPluginExtension.kt
```diff
@@ -0,0 +1,27 @@
+package de.triplet.gradle.play
+
+import java.io.File
+
+open class PlayPublisherPluginExtension {
+    var serviceAccountEmail: String? = null
```
ChristianBecker: Should we deprecate those?

src/main/kotlin/de/triplet/gradle/play/PlayPublishTask.kt
```diff
@@ -0,0 +1,46 @@
...
+    var playAccountConfig: PlayAccountConfig? = null
+
+    var service: AndroidPublisher? = null
+
+    fun publish() {
+        if (service == null) {
+            service = AndroidPublisherHelper.init(extension, playAccountConfig)
+        }
+
+        edits = service!!.edits()
```
ChristianBecker: I don't like this. Unfortunately we can't get rid of this operator as `service` is a mutable property :(

src/main/kotlin/de/triplet/gradle/play/AndroidPublisherHelper.kt
bhurling: Sure. Can be moved. The helper is using a `JacksonFactory` and a `HttpTransport` which should be mockable in the future

src/main/kotlin/de/triplet/gradle/play/PlayPublishListingTask.kt
bhurling: From what I see down below it actually fails with an error instead of logging a warning. Is that the desired behavior?

src/main/kotlin/de/triplet/gradle/play/PlayPublisherPluginExtension.kt
bhurling: Yes. Using the old approach should trigger a warning and refer to the new playAccountConfig

src/main/kotlin/de/triplet/gradle/play/PlayPublishListingTask.kt
ChristianBecker: As far as I can tell it uses `logger.error(..)` to just print a message to the logger (see line 144). This is the old/existing behavior.

bhurling: True

ChristianBecker: We can (should?) however change this into an error. In the same change we should move the validations ahead of any executions of the Play Publisher API, so that the executions/updates are "atomic". See #167 as a possible integration.

.gitignore
```diff
@@ -1,6 +1,5 @@
-build
 .idea
-*.iml
```
rocboronat: Warning! Ignoring the `.iml` files is a good idea

ChristianBecker: You're right. My bad. I will revert this line! Thanks.

rocboronat: You're welcome! 👍

README.md
```diff
@@ -47,7 +47,7 @@ buildscript {
 
     dependencies {
     	// ...
-        classpath 'com.github.triplet.gradle:play-publisher:1.2.1'
+        classpath 'com.github.triplet.gradle:play-publisher:1.2.2-beta1'
```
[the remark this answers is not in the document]
SUPERCILEX: Agreed. Also, I think we should start development on v2.0 or v1.3 so we can bump the AGP version, meaning 1.2.2 won't be correct.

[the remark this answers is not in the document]
gtcompscientist: I'm ok with both of those thoughts. I suggest that the Kotlin version be 1.2.3, then we can branch that to start working on v2.0 using the latest AGP version (including Bundle prep)

Because, if we don't merge this back to master, PRs incoming will continue to be on the Groovy files. :/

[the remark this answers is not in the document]
SUPERCILEX: I agree with you that this PR needs to be merged before we start working on v2.0, but I don't think we should release v1.2.2: if the release is just moving to Kotlin, we'll be adding bugs, not removing them. 😂

PR #203, opened 2017-03-23
Use project.extensions.getByType(AppExtension.class)

ChristianBecker:
… instead of `project.android`. This will increase IDE support as we now know the type.


bhurling:
Looks good

PR #212, opened 2017-05-11
Allow overriding rootUrl.

LegNeato:
This is useful for specifying a proxy or for pointing to
localhost for testing.


LegNeato:
Any thoughts on this PR?

SUPERCILEX:
We've decided against allowing customization of the network stack. However, if you have a compelling case, please don't hesitate to open an issue so we can discuss it further.

gtcompscientist:
As an aside, you could always add a local reroute to your machine.

Issue #216, opened 2017-05-23
publish task fails on gradle 4.0 and gradle plugin 3.0.0.-alpha1


dannyroa:
I'm getting the following error after I upgraded gradle to 4.0-milestone and gradle plugin to 3.0.0.-alpha1.

Looks like related to https://issuetracker.google.com/issues/38408231

```
Caused by: java.lang.RuntimeException: Not valid.
        at com.android.ide.common.build.ApkData.getMainOutputFile(ApkData.java:136)
        at com.android.build.gradle.internal.api.BaseVariantOutputImpl.getOutputFile(BaseVariantOutputImpl.java:60)
        at com.android.build.gradle.internal.api.ApkVariantOutputImpl_Decorated.getOutputFile(Unknown Source)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.getProperty(BeanDynamicObject.java:228)
        at org.gradle.internal.metaobject.BeanDynamicObject.tryGetProperty(BeanDynamicObject.java:171)
        at org.gradle.internal.metaobject.CompositeDynamicObject.tryGetProperty(CompositeDynamicObject.java:55)
        at org.gradle.internal.metaobject.AbstractDynamicObject.getProperty(AbstractDynamicObject.java:59)
```

NiKoTron:
fixed in gradle plugin to 3.0.0.-alpha**2**

cwliu:
I also got this error when I upgrade to Android Studio 3.0 Canary 2. The following is my workaround: 

1. Go to `File > Project Structure`. Click `Project` section
2. Downgrade `Gradle version` to `3.3` and `Android plugin version` to `2.3.2`

dannyroa:
@cwliu: That doesn't work for us because we already made changes specific to the latest gradle plugin. We're also running this on CI.

OrdonTeam:
@dannyroa 
Here is change set that fixed same issue in our project:
https://github.com/elpassion/el-space-android/commit/f2ac5a7dd1c0c176e3444e7e87b8c851e994638a

cypressious:
@OrdonTeam Can you create a pull request with the fix?

OrdonTeam:
Use build tools gradle plugin 3.0.0-alpha4.

ChristianBecker:
The current version of the plugin is properly working on 3.x.y

PR #223, opened 2017-06-09
Migrate to gradle 4.0 and build tools 3.0.0-alpha4

OrdonTeam:


OrdonTeam:
This resolves #216

src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
```diff
@@ -19,10 +19,10 @@ class PlayPublishApkTask extends PlayPublishTask {
     @TaskAction
     publishApks() {
         publish()
-
+        def apkDir = new File(inputFolder.getParentFile().getParentFile(), "apk/${variant.dirName}")
```
OrdonTeam: I'm it can be done in better way :/

ChristianBecker: On a short note without proper review this can be written as:
```
def apkDir = new File(inputFolder.parentFile.parentFile, "apk/${variant.dirName}")
```

ChristianBecker:
Thanks, I'll try to perform a proper review on the next days. We must be careful that our solution is backwards-compatible for the 2.x plugin.

cypressious:
@ChristianBecker As the android plugin 3.0.0 is not backward compatible itself, maybe it makes sense to release this as a new major version without backward compatibility guarantees.

src/main/groovy/de/triplet/gradle/play/PlayPublishApkTask.groovy
OrdonTeam: Improved in fd17d2d2027a2fb95ceba12a1558598f626cf14d

OrdonTeam:
As I checked, this solution is not backwards compatible.

ChristianBecker:
@bhurling : Any opinions on this? Should we release a `2.x.y` without backwards compatibility or should we try to support both `1.x.y` and `2.x.y` with one version?

bhurling:
I think we can drop support with a major version bump.

OrdonTeam:
I was trying to support both versions but I haven't found other solution than try catching missing property exception. I'm voting for dropping support.

nhaarman:
Is there any ETA on this? Our project's build update is stuck because of this issue.

OrdonTeam:
@nhaarman meanwhile you can apply similar changes as here: https://github.com/elpassion/el-space-android/commit/f2ac5a7dd1c0c176e3444e7e87b8c851e994638a
There is compiled plugin migrated as in PR.
Ps. Love mockito kotlin

nhaarman:
I retried the plugin with `3.0.0-alpha4`, and now the error does not occur anymore. Can anyone confirm?

OrdonTeam:
You are right. 1.2.0 works just fine with build tool gradle 3.0.0-alpha4

OrdonTeam:
After changing to build tools gradle `3.0.0-alpha4` I were able to restore previous implementation of `PlayPublishApkTask`.
Now in this PR only test code had to be changed.

nhaarman:
Well, it might be broken again in `alpha5`. We'll just have to wait and see I guess.

OrdonTeam:
Don't think so. Whole issue was because they deprecated one method but they were still using it in their classes. It was obviously broken on their side. I have looked thought the code of `alpha4` and it looks like they want to preserve `ApkVariantOutput` api.
Any way you are right that they can break it again.

ChristianBecker:
@OrdonTeam : Can you rebase, update to latest Gradle wrapper & plugin, so that we can get this merged? I think we're save to drop support for pre-3.0 plugins.

OrdonTeam:
No I cannot. At least not soon... Feel free to take ownership over this PR

ChristianBecker:
Superseded by #256

OrdonTeam:
Thanks

PR #241, opened 2017-10-12
Update the Gradle wrapper and Android Gradle plugin

scottkennedy:
4.2.1 and 2.3.3 are the latest stable versions.


hannesa2:
now 3.0.0 is the way to go

ChristianBecker:
@scottkennedy : Thanks, but I think we should wait for #223

Issue #168, opened 2016-09-16
Upload only if Local App version is higher than app version in Play Store


tridib:
I want to automate the App upload to Play store through Jenkins. I was wondering if this plugin can upload only if there is a version upgrade. Right publish tasks fail if version is same. If there a way to configure the plugin to check the version in play store and upload if current version is higher than that of play store?

SUPERCILEX:
Question for the community: should not uploading the APK if there's a conflict be the default or should automatically upgrading the version code be the default?

gtcompscientist:
I would say not uploading is better. And how the task is managed after that is what becomes important. But, it could easily be a configuration
Options of:
- fail task (default)
- warn (log a warning for the user to see, but continue)
- ignore (skip the task entirely)

gtcompscientist:
As a follow-up, the idea of a warning, with no upload would have a good use case for batching uploads where something fails halfway through (lost internet, whatever) and you have to restart.

PR #248, opened 2018-02-07
Feature: Automatically increment android versionCode.

sebokopter:
For publishing our Apps in the PlayStore we use the VersionCode from the publisher, increment and add it to the APKs for every release.
Since we use it heavily in our company others might also be interested in this feature.


SUPERCILEX:
Hi there, we've moved to Kotlin so this PR is unfortunately obsolete. Sorry about that! 😊 See #168 for discussion. And feel free to open a new PR in Kotlin if you have the chance!

PR #261, opened 2018-05-08
Make 'internal' track default

PiotrWpl:
Maybe now "Internal" track should be default, instead of alpha?


SUPERCILEX:
The internal track maxes out at 100 users and seems really closed off. I feel like it makes sense for the alpha track to be the default since it already requires users to be invited to a Google+ community or a group. @bhurling what do you think?

bhurling:
It would be a breaking change because some users might be using the plugin with no track specified and expect their apks to show up in the alpha track. Apart from that I don't really have a preference. Seems reasonable to push to the lowest impact track by default.

@PiotrWpl could you adjust the tests?

bhurling:
Closed in favor of #274

bhurling:
We have rewritten the plugin in kotlin. But we should not forget about this one

Issue #259, opened 2018-04-19
Is it possible to release to the 'Internal test' track?


HarshMindera:

huytran0605:
same question

burntcookie90:
https://github.com/Triple-T/gradle-play-publisher/commit/4bb0bd9203c3337affffa9ae95d971826ec910c2 looks like this was merged recently, so yes.

fearlessforhim:
Any ETA on when this change goes out?

jonneymendoza:
Any info on how to build an internel release using 1.2.1-beta1?

bhurling:
Just set the track to internal

```
play {
    track = 'internal'
}
```

thekekc:
When i checkout 1.2.1-beta1 version and try to build project i get this error:"minimum supported Gradle vesion is 4.4. Current is 4.1".  Still works with 1.2.0.

bhurling:
@ChristianBecker This is probably due to updating our android plugin dependency to 3.1.0 in #256 Apparently this change is not backwards compatible. Might make sense to temporarily revert that change for a proper version 1.2.1.

bhurling:
Fixed in 1.2.1

PR #264, opened 2018-05-22
Downgrade android plugin to 2.3.0 to not break other people's builds

bhurling:
This change ensures backwards compatibility in case people are still using a version of Gradle lower than 4.4 (See comment https://github.com/Triple-T/gradle-play-publisher/issues/259#issuecomment-390972786). You can find a list of compatible versions at https://developer.android.com/studio/releases/gradle-plugin#behavior_changes_2.

With that change we can now safely release a version 1.2.1. Bumping the android plugin version should be considered a breaking change.

cc @ChristianBecker

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ repositories {
     google()
 }
 dependencies {
-    compile 'com.android.tools.build:gradle:3.1.0'
+    compile 'com.android.tools.build:gradle:2.3.0'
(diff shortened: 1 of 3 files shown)


bhurling:
A better solution might be to set `compileOnly` for `com.android.tools.build:gradle:x.y.z` so we wouldn't clash with the version referenced by the app project.

Issue #268, opened 2018-05-24
Revamp logging and validation


SUPERCILEX:
This issue coalesces multiple requests related to logging. We need to:
- validate signing configs
- log what's being uploaded
- show progress while uploading
- clarify error messages
- ensure important logs show up in the build without having to pass a `--info` flag

PR #269, opened 2018-05-24
Revamp README

SUPERCILEX:


README.md
```diff
@@ -1,217 +1,212 @@
...
+Once that's done, you'll need to grant the following permissions to your service account for this
+plugin to work (go to Settings -> Developer account -> API access -> Service Accounts):
 
 ![permissions.png](https://cloud.githubusercontent.com/assets/242983/17809992/95ea4eaa-661a-11e6-9879-521df4f14735.png)
 
-Once you finished the setup you have a so called *service account email address* and a *p12 key file* that we will use later on.
-
-### Signing Configuration
-
-Please make sure to assign a valid signing configuration to your release build type. Otherwise, there won't be a publishable (signed) APK. In that case, the plugin won't create any of its tasks.
```
bhurling: I'd like to keep a hint for that as long as we're not changing the behavior. This has been a constant source of error when setting up the plugin

README.md
```diff
@@ -1,217 +1,212 @@
...
-
-In case you are using product flavors you will get one of the above tasks for every flavor. E.g. `publishApkPaidRelease` or `publishListingPaidRelease`.
+Should you choose to use product flavors, there will be an apporpriately named task for each flavor.
+E.g. `publishApkPaidRelease` or `publishListingPaidRelease`.
 
-## Authentication
+## Authenticating the plugin
 
-Similar to Android's `signingConfigs` you need to setup so called `playAccountConfigs` to authorize your requests. Drop in your service account email address and the p12 key file you generated in the API Console here.
+After you've gone though the [Service Account setup](#service-account), you should have a JSON file
```
bhurling: typo: through

README.md
```diff
@@ -1,217 +1,212 @@
...
-    untrackOld = true // will untrack 'alpha' while upload to 'beta'
+    untrackOld = true // Will untrack 'alpha' if needed to upload to 'beta'
 }
 ```
 
-This will untrack whatever versions are currently blocking the publishing process. That is: every APK with a version lower than the one being uploaded in any of the tracks lower than specified by `track`.
+`untrackOld` will untrack all versions blocking a publication: that is, every APK with a version
+code lower than the one being uploaded _and_ a lower track will be untracked. Example: publishing an
+APK with version code 42 to the beta channel will untrack versions 41 and lower from alpha and
+above. However, it will not untrack the stable channel or any alpha and above chanel with version
```
bhurling: "and above" seems wrong here. What is it referring to?

README.md
```diff
@@ -1,217 +1,212 @@
...
-        }
-    }
-    
-    defaultConfig {
-        // ...
-        
-        playAccountConfig = playAccountConfigs.defaultAccountConfig
-    }
+play {
+    jsonFile = file('your-key.json')
```
bhurling: Should this be the preferred way of authentication for single flavor projects? I remember that we actually wanted to deprecate that solution in favor of `playAccountConfigs`

bhurling: Also, we should keep a reference to pk12 files somewhere (maybe in the advanced section) as long as Google allows to authenticate that way. We probably have to keep that mechanism forever to not force users to recreate service accounts

README.md
SUPERCILEX: You're right. In the future, it'd be nice to just break the build with a pretty error message, see #268.

README.md
SUPERCILEX: I was referring to the `internal` track, but you're right, it's just confusing. Removing it.

README.md
```diff
@@ -1,217 +1,230 @@
...
-    
-    defaultConfig {
-        // ...
-        
-        playAccountConfig = playAccountConfigs.defaultAccountConfig
-    }
+android { ... }
+
+play {
+    jsonFile = file('your-key.json')
```
SUPERCILEX: > Should this be the preferred way of authentication for single flavor projects? I remember that we actually wanted to deprecate that solution in favor of `playAccountConfigs`.

Personally, I like simplicity of this method. However, if you think we should kill it in v2.0 to unify authentication, I'm all for that. It's up to you, I'm fine either way. 😊

PR #249, opened 2018-02-21
Create "All" tasks to upload all release variants all at once.

sebokopter:
As a app can have a lot of different variants it can be getting quite a lot of gradle tasks to list to publish them all simulatiously.
e.g. ./gradlew publishFreeRelease publishPaidRelease publishDemoRelease publishSubscriptionRelease ...

To make this more convient on the library consumer side this commit adds the tasks "publishAll", "publishApkAll" and "publishListingAll" which does the same as "publish*Release", "publishApk*Release" and "publishListing*Release" respectively but for all variants.

Issue #181, opened 2016-12-16
Add support for in-app products


Pitel:
We'd like to automate publishing localized in-app products.

API: https://developers.google.com/android-publisher/api-ref/inappproducts

Issue #117, opened 2016-01-07
Publish all productFlavor


aat-antoine:
I know

> In case you are using product flavors you will get one of the above tasks for every flavor. E.g. publishApkPaidRelease or publishListingPaidRelease

But it will be nice if there was a task `publishAllApksRelease` to publish all apk for every flavor. (same for listing)

pablisco:
This should give you what you want ;)
```
task('publishAllApksRelease').dependsOn(tasks.findAll { it.name.startsWith("publishApk") })
```

siloebb:
@pablisco don't work this method.
Anyone know what we need do?

brp-david:
This is what I use to update all my 70+ flavours.
It should give exactly the same result as pabliscos suggestion, but I can verify that mine works:

```
task updateBleedingEdge(group: 'util') << {}
updateBleedingEdge.dependsOn {
    tasks.findAll { it.name.startsWith('publishApk') }
}
```

johnnyzen:
Hi

The above works great, but is there a way to ignore the below error (as its the latest version) and go to the next publishApk task?

Error:Execution failed for task ':app:publishApkArsenalRelease'.
> com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
  {
    "code" : 403,
    "errors" : [ {
      "domain" : "androidpublisher",
      "message" : "APK specifies a version code that has already been used.",
      "reason" : "apkUpgradeVersionConflict"
    } ],
    "message" : "APK specifies a version code that has already been used."
  }

yottabrain:
Just renamed task given by @brp-david so it will come under 'play store' group along with other apkrelease tasks

```
task publishApkAllRelease(group: 'play store') << {}
publishApkAllRelease.dependsOn {
    tasks.findAll { it.name.startsWith('publishApk') }
}
```

nielsdelporte:
Does the plugin support variations in listings/metadata depending on product flavour?
Say you want to have a different set of screenshots for free/premium flavours, or different descriptions.

PR #273, opened 2018-05-25
All flavors

gtcompscientist:
Adds "All" level tasks for each of the tasks currently supported. Used #249 as the basis.

#117 is covered in this

This will also lays the groundwork for any other "all tasks", such as:
- Promoting builds to different channels
- Adding/updating in-app products #181


ChristianBecker:
Let's get it merged!

PR #288, opened 2018-06-02
add maven publish plugin

bhurling:


bhurling:
Working fine with the defaults

Issue #284, opened 2018-05-28
Add `modifyRelease[Flavor]` task


gtcompscientist:
Add a new task to promote releases. Looking for input on the task and manner of promotion.

My suggestion:
- Command-line input would be `fromTrack` and `toTrack` [optional]
- `toTrack` must be the same or "higher"
- Defaults to the plugin configured `status` of a release
- Releases status can also be overridden with any of the possible statuses

Available statuses:
- `completed`
- `draft`
- `halted`
- `inProgress` w/`rollout` %

The `modifyRelease` task would also have an "all" task for all `variants`.

Would solve #104, coincide with #252 and #274. And probably respect whatever logic is used for #125.

SUPERCILEX:
Making some internal discussion public: it might be better to add a builder to the `play` block so that devs can choose where their values come from:
```kotlin
play {
    // ...
    releaseModifiers {
        fromTrack = properties["fromTrack"] // Could also be hardcoded, retrieved from the network, or even randomly generated to create an "I'm feeling lucky" button 😁
        toTrack = properties["toTrack"]
        status = properties["status"]
    }
}
```

Then the same way of passing args via the CLI would work:
```sh
./gradlew modifyReleasePaid -PfromTrack=internal -Pstatus=completed | ./gradlew modifyReleasePaid -PfromTrack=production -Pstatus=draft
```

gtcompscientist:
Also allows for easier unit testing.

gtcompscientist:
More discussion:
- Should "What's New" be updated from disk? Or should that be only for publishing?
- When updating, what should happen to the old Track? (Just set to empty?)
- Should we allow naming of Track Releases, to allow for multiple, and/or promoting a single release?
- If allowing multiple, what do we put in the new Track, the promoted release? Search/replace based on name? Add to what exists?

Multiple Track Releases would require research on how the API works with those. (What tracks can have multiple?)

gtcompscientist:
Named releasees are specifically for Alpha channel, from my research, and they MUST be named. So we'll need a way to distinguish them, and know which track to modify.

Issue #252, opened 2018-03-01
Upload APK to make a draft?


vanlooverenkoen:
Is it possible to upload the apk to the play console but only to make a draft?

bhurling:
Sounds interesting. I'm not sure if Google's API allows it. If it does we could add a configuration flag

gtcompscientist:
It looks like the Tracks API which requires a change to the "What's New" data includes this ability:
https://developers.google.com/android-publisher/api-ref/edits/tracks

<img width="430" alt="screen shot 2018-05-18 at 2 51 48 pm" src="https://user-images.githubusercontent.com/230875/40252655-0409b5a2-5aab-11e8-9014-7a3bc627c213.png">

bhurling:
Fixed via #291

PR #291, opened 2018-06-02
V3 upgrade

gtcompscientist:
Upgrades to v3 of the Publisher API, with support for drafts.  #283 #252

Also moves the `updateTrack` logic in preparation of the Modify tasks. #284

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 
 dependencies {
     implementation("com.android.tools.build:gradle:3.0.1")
-    implementation("com.google.apis:google-api-services-androidpublisher:v2-rev77-1.23.0") {
+    implementation("com.google.apis:google-api-services-androidpublisher:v3-rev4-1.23.0") {
         exclude("com.google.guava", "guava-jdk5") // Remove when upgrading to AGP 3.1+
     }
     implementation(kotlin("stdlib-jdk7"))
--- a/plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
+++ b/plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
@@ -1,10 +1,7 @@
 package de.triplet.gradle.play
 
 import com.google.api.services.androidpublisher.AndroidPublisher
-import de.triplet.gradle.play.internal.ImageType
-import de.triplet.gradle.play.internal.LISTING_PATH
-import de.triplet.gradle.play.internal.ListingDetail
-import de.triplet.gradle.play.internal.PlayPublishTaskBase
+import de.triplet.gradle.play.internal.*
 import org.gradle.api.tasks.TaskAction
 import java.io.File
 import java.net.URL
@@ -59,19 +56,16 @@ open class BootstrapTask : PlayPublishTaskBase() {
     }
 
     private fun AndroidPublisher.Edits.bootstrapWhatsNew(editId: String) {
-        val versionCode = apks()
+        val maxTrack = tracks()
                 .list(variant.applicationId, editId)
-                .execute().apks
-                ?.map { it.versionCode }
-                ?.max() ?: return
-        val apkListings = apklistings()
-                .list(variant.applicationId, editId, versionCode)
-                .execute()
-                .listings ?: return
+                .execute().tracks
+                ?.maxBy { TrackType.valueOf(it.track) } ?: return
 
-        for (listing in apkListings) {
-            File(outputFolder, "${listing.language}/${ListingDetail.WHATS_NEW.fileName}")
-                    .writeText(listing.recentChanges)
+        maxTrack.releases
+                .maxBy { it.versionCodes.max() ?: Long.MIN_VALUE }
+                ?.releaseNotes?.forEach {
+            File(outputFolder, "${it.language}/${ListingDetail.WHATS_NEW.fileName}")
+                    .writeText(it.text)
         }
     }
 
--- a/plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishTaskBase.kt
+++ b/plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishTaskBase.kt
@@ -6,8 +6,12 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.androidpublisher.AndroidPublisher
 import com.google.api.services.androidpublisher.AndroidPublisherScopes
+import com.google.api.services.androidpublisher.model.LocalizedText
+import com.google.api.services.androidpublisher.model.Track
+import com.google.api.services.androidpublisher.model.TrackRelease
 import de.triplet.gradle.play.PlayPublisherExtension
 import org.gradle.api.DefaultTask
+import java.io.File
 
 abstract class PlayPublishTaskBase : DefaultTask() {
     lateinit var extension: PlayPublisherExtension
@@ -71,5 +75,47 @@ abstract class PlayPublishTaskBase : DefaultTask() {
     ) = read {
         block(it)
         commit(variant.applicationId, it).execute()
+    }
+
+    internal fun AndroidPublisher.Edits.updateTracks(
+            editId: String,
+            inputFolder: File,
+            releaseStatus: String,
+            versions: List<Long>,
+            trackType: String,
+            userPercent: Double) {
+        val track = tracks()
+                .list(variant.applicationId, editId)
+                .execute().tracks?.firstOrNull { it.track == trackType }
+                ?: Track()
+
+        var releaseText: List<LocalizedText>? = null
+        if (inputFolder.exists()) {
+            releaseText = inputFolder.listFiles(LocaleFileFilter).map { locale ->
+                val fileName = ListingDetail.WHATS_NEW.fileName
+                val file = run {
+                    var file = File(locale, "$fileName-${extension.track}").orNull()
+                    if (file == null) file = File(locale, fileName).orNull()
+                    file
+                } ?: return@map null
+                val recentChanges = File(file, fileName).readProcessed(
+                        ListingDetail.WHATS_NEW.maxLength,
+                        extension.errorOnSizeLimit
+                )
+                LocalizedText().setLanguage(locale.name).setText(recentChanges)
+            }.filterNotNull()
+        }
+        val trackRelease = TrackRelease().apply {
+            releaseNotes = releaseText
+            status = releaseStatus
+            userFraction = if (status == ReleaseStatus.INPROGRESS.status) userPercent else 0.0
+            versionCodes = versions
+        }
+
+        track.releases = listOf(trackRelease)
+
+        tracks()
+                .update(variant.applicationId, editId, trackType, track)
+                .execute()
     }
 }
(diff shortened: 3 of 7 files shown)


plugin/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
 
 dependencies {
     implementation("com.android.tools.build:gradle:3.0.1")
-    implementation("com.google.apis:google-api-services-androidpublisher:v2-rev77-1.23.0") {
+    implementation("com.google.apis:google-api-services-androidpublisher:v3-rev4-1.23.0") {
```
ChristianBecker: There's already `v3-rev5-1.23.0` available

plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
```diff
@@ -1,10 +1,7 @@
 package de.triplet.gradle.play
 
 import com.google.api.services.androidpublisher.AndroidPublisher
-import de.triplet.gradle.play.internal.ImageType
-import de.triplet.gradle.play.internal.LISTING_PATH
-import de.triplet.gradle.play.internal.ListingDetail
-import de.triplet.gradle.play.internal.PlayPublishTaskBase
+import de.triplet.gradle.play.internal.*
```
ChristianBecker: We should forbid star imports

plugin/build.gradle.kts
gtcompscientist: @ChristianBecker They're already up to `rev6`! 😆

plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
```diff
@@ -59,19 +60,16 @@ open class BootstrapTask : PlayPublishTaskBase() {
...
-        val apkListings = apklistings()
-                .list(variant.applicationId, editId, versionCode)
-                .execute()
-                .listings ?: return
-
-        for (listing in apkListings) {
-            File(outputFolder, "${listing.language}/${ListingDetail.WHATS_NEW.fileName}")
-                    .writeText(listing.recentChanges)
+                .execute().tracks
+                ?.maxBy { TrackType.valueOf(it.track) } ?: return
```
SUPERCILEX: Shouldn't this use `fromString`?

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishTaskBase.kt
```diff
@@ -72,4 +76,46 @@ abstract class PlayPublishTaskBase : DefaultTask() {
         block(it)
         commit(variant.applicationId, it).execute()
     }
+
+    internal fun AndroidPublisher.Edits.updateTracks(
```
SUPERCILEX: This is only used in the `PublishApkTask`, why are we moving it to the base?

plugin/src/main/kotlin/de/triplet/gradle/play/internal/Tracks.kt
```diff
@@ -1,12 +1,26 @@
 package de.triplet.gradle.play.internal
 
-internal val Track.superiors get() = Track.values().takeWhile { it != this && it != Track.ROLLOUT }
+internal val TrackType.superiors get() = TrackType.values().takeWhile { it != this && it != TrackType.ROLLOUT }
 
-internal enum class Track(val publishedName: String) {
+internal enum class TrackType(val publishedName: String) {
```
SUPERCILEX: Renamed to `TrackType` to not conflict with Google `Track`?

plugin/src/main/kotlin/de/triplet/gradle/play/internal/Tracks.kt
```diff
@@ -1,12 +1,26 @@
...
     INTERNAL("internal"),
     ALPHA("alpha"),
     BETA("beta"),
     ROLLOUT("rollout"),
-    PRODUCTION("production")
+    PRODUCTION("production");
+
+    companion object {
+        fun fromString(value: String): TrackType {
+            return values().first { it.publishedName.equals(value, true) }
```
SUPERCILEX: Can we make this an expression?

plugin/src/main/kotlin/de/triplet/gradle/play/internal/Tracks.kt
```diff
@@ -1,12 +1,26 @@
...
+    }
+}
+
+internal enum class ReleaseStatus(val status: String) {
+    COMPLETED("completed"),
+    DRAFT("draft"),
+    HALTED("halted"),
+    INPROGRESS("inProgress");
 }
+
```
SUPERCILEX: Nit: unnecessary new line

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishTaskBase.kt
gtcompscientist: That's there because the `modify` work will use it as well.

plugin/src/main/kotlin/de/triplet/gradle/play/internal/Tracks.kt
gtcompscientist: Yep, got tired of fully-qualified classes here and there.

plugin/src/main/kotlin/de/triplet/gradle/play/internal/Tracks.kt
gtcompscientist: 👍 I actually thought I discarded that line. 👎

plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
```diff
@@ -86,4 +86,4 @@ open class BootstrapTask : PlayPublishTaskBase() {
 
     private fun String.write(dir: File, detail: ListingDetail) =
             File(dir, detail.fileName).writeText(this)
-}
+}
```
SUPERCILEX: @gtcompscientist Lol, now we don't have a new line. Can we add 1 back here and in `Tracks.kt`. I recommend going into IntelliJ/AS `Settings -> Editor -> General`. Under `Other`, you'll find `Ensure line feed at file end on save`.

gtcompscientist: Hahaha, yeah.

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishTaskBase.kt
gtcompscientist: That works for me, I'll refactor when I do the `modify` work.

SUPERCILEX: ❓ Can't you do it now? Or at least move this back to where it was in the `PublishApkTask`? If you want, I'd be happy to do that for you. 😄

plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
```diff
@@ -59,19 +60,16 @@ open class BootstrapTask : PlayPublishTaskBase() {
...
-                .execute()
-                .listings ?: return
-
-        for (listing in apkListings) {
-            File(outputFolder, "${listing.language}/${ListingDetail.WHATS_NEW.fileName}")
-                    .writeText(listing.recentChanges)
+                .execute().tracks
+                ?.maxBy { TrackType.fromString(it.track) } ?: return
+
+        maxTrack.releases
```
bhurling: could we just chain the `releases` with another `?.` operator? That way we can avoid the `val maxTrack` and that elvis operator in L66 that doesn't assign anything but only returns from the method.

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishPackageBase.kt
```diff
@@ -0,0 +1,52 @@
...
+    ) {
+        val track = tracks()
+                .list(variant.applicationId, editId)
+                .execute().tracks?.firstOrNull { it.track == trackType } ?: Track()
+
+        val releaseText = if (inputFolder.exists()) {
+            inputFolder.listFiles(LocaleFileFilter).mapNotNull { locale ->
+                val fileName = ListingDetail.WHATS_NEW.fileName
+                val file = run {
+                    var file = File(locale, "$fileName-${extension.track}").orNull()
```
bhurling: Could we write this without a var? Like so:
```
run {
    File(locale, "$fileName-${extension.track}").orNull() ?: File(locale, fileName).orNull()
}

File: plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishPackageBase.kt [hunk 11]
```diff
@@ -0,0 +1,52 @@
...
+                )
+                LocalizedText().setLanguage(locale.name).setText(recentChanges)
+            }
+        } else {
+            null
+        }
+        val trackRelease = TrackRelease().apply {
+            releaseNotes = releaseText
+            status = releaseStatus
+            userFraction = if (status == ReleaseStatus.INPROGRESS.status) userPercent else 0.0
```
bhurling: does this mean users have to set both `userFraction` _and_ `releaseStatus` in the `play` block? That would be a breaking change we would have to explain in a migration doc

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishPackageBase.kt
```diff
@@ -0,0 +1,52 @@
...
+            releaseStatus: String,
+            versions: List<Long>,
+            trackType: String,
+            userPercent: Double
+    ) {
+        val track = tracks()
+                .list(variant.applicationId, editId)
+                .execute().tracks?.firstOrNull { it.track == trackType } ?: Track()
+
+        val releaseText = if (inputFolder.exists()) {
```
bhurling: Let's use plural here because this holds a list of LocalizedText objects, right?

plugin/src/main/kotlin/de/triplet/gradle/play/BootstrapTask.kt
gtcompscientist: I can actually make it an expression, with that change, that's fine.

plugin/src/main/kotlin/de/triplet/gradle/play/internal/PlayPublishPackageBase.kt
gtcompscientist: I guess I'm not understanding. It should have the same functionality as today. If the status isn't set, it defaults to `completed`, if the status is set to `rollout`, then the `userPercent` is used. Neither is required to be set by the user.

bhurling: We have three parameters now: track, status and userPercent. 1.x only had track and userPercent. So current projects that use stayed rollouts are configured like this:

- track: rollout
- userPercent: e.g. 0.05
- no value for status because it didn't exist, thus: defaults to completed

That means we would silently override their userPercent to 0

gtcompscientist: I see what you're saying now. From the API docs. We need to send either `inProgress` or `halted` if we set the release to `rollout`. How about if `track` is `rollout`, then we default to `inProgress`, with overrides still available, and an error if they set `rollout` and an incompatible `status`.

PR #295, opened 2018-06-03
Simplify and standardize validation

SUPERCILEX:
Our current validation depends on the dev setting both fields or implicitly sets them otherwise. This PR moves validation to `afterEvaluate`, keeps our extension free of logic, and requires explicitly setting the release status for clarity.


plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublisherPluginTest.groovy
```diff
@@ -186,41 +177,28 @@ class PlayPublisherPluginTest {
...
-                evaluateProjectWithTrack(TrackType.ROLLOUT).extensions.findByName('play').releaseStatus)
+        assertTrue(evaluateProjectWithTrackAndStatus(TrackType.ROLLOUT, ReleaseStatus.IN_PROGRESS))
     }
 
-    private evaluateProjectWithTrack(TrackType trackType) {
+    private evaluateProjectWithTrackAndStatus(TrackType trackType, ReleaseStatus status) {
         def project = TestHelper.evaluatableProject()
         project.play {
             track trackType.publishedName
+            releaseStatus status.publishedName
```
bhurling: The missing `releaseStatus` was the whole point to ensure backwards compatibility. Users should not have to set it and we should always provide sane defaults as it is written in the dev-docs. If users want to publish to `rollout` track we should default to `inProgress`. Otherwise (internal, alpha, beta, production), we should default to `completed`

SUPERCILEX: Gotya, that makes sense.

PR #296, opened 2018-06-03
Output override for using APKs out of the standard build process

### LINKED ISSUE #360: Allow providing artifact to upload instead of building from source
We store our RC builds on a server. When a build reaches QA approval, we'd like to upload that specific build. Our CI executes a job that will run publishApk task on demand on a clean directory. Placing a pre-built APK in the respective `build/outputs` folder doesn't seem to make publishApk task to take the APK and upload it. Instead, it goes through the whole build process from scratch. 
Is it possible to make publishApk task upload existing APK instead of building from source?

gtcompscientist:
Opening this now to discuss this option further. Especially the name of the option, and the README.

Fixes #360


SUPERCILEX:
I'm just confused as to why the tool can't output the APK right back into the original build folder... am I missing something?

gtcompscientist:
The `publish` task relies on the `assemble` task, meaning it will **always** overwrite that just before publishing.

README.md
```diff
@@ -25,6 +25,8 @@ the Google Play Store from a continuous integration server or anywhere you have
    1. [Using multiple Service Accounts](#using-multiple-service-accounts)
    1. [Encrypting Service Account keys](#encrypting-service-account-keys)
    1. [Running custom tasks before publishing](#running-custom-tasks-before-publishing)
+1. [Advanced](#advanced)
+   1. [Using Alternate Build Folder](#alternate-build-folder)
```
SUPERCILEX: `Using alternate build folders` with the `s` and no caps since we're using sentence case

README.md
```diff
@@ -364,3 +366,22 @@ project.afterEvaluate {
...
+## Advanced
+
+Options and configurations that are unusual, but available for advanced use-cases.
+
+### Alternate Build Folder
+
+If you use an external tool to process your APK file, you will need to override the standard
+publish process. With this enabled, the standard output will be ignored, and any `.apk` files
+in the provided folder will uploaded instead. If you have multiple flavors with different
+`applicationId`s you will need to use a property and override for each application you are
```
SUPERCILEX: I don't understand this sentence. There's only one property, what do you mean you need to set it for each?

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -82,4 +83,10 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
                 _track = TrackType.ROLLOUT
             releaseTrackSet = true
         }
+
+    /**
+     * Optional folder to find APK files. Only for use when the APK is built by an external process
+     */
+    var overrideBuildOutput: File? = null
```
SUPERCILEX: This property name is confusing. It makes it sound like it's _our_ output. `apkInputFolder` makes more sense.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -82,4 +83,10 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
...
+
```
SUPERCILEX: Nit: uncessassary line

plugin/src/main/kotlin/com/github/triplet/gradle/play/PublishApkTask.kt
```diff
@@ -16,8 +16,9 @@ open class PublishApkTask : PlayPublishPackageBase() {
 
     @TaskAction
     fun publishApks() = write { editId: String ->
-        //TODO: If we take in a folder here as an option, we can fix #233, #227
-        val publishedApks = publishApks(editId)
+        val files = extension.overrideBuildOutput?.listFiles({ _, name -> name.endsWith(".apk") })?.toList()
```
SUPERCILEX: The filter should be extracted into an object in a new `Packaging.kt` file:
https://github.com/Triple-T/gradle-play-publisher/blob/591a1138a5e602bfc3d5377776c35d3c33d79e98/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Listings.kt#L13-L15

SUPERCILEX:
I don't think that's true about it _always_ overwriting the folder since you could just add a task that depends on assemble and processes the APK from there. However, I can see how someone would want to keep both the unprocessed and processed APKs so this sounds like a good idea anyway.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
gtcompscientist: I think it should be more generic than that. But agree on the `input` vs. `output`

SUPERCILEX:
@gtcompscientist I don't think you're merging things correctly... Or GitHub is broken

gtcompscientist:
@SUPERCILEX I honestly don't know what happened there, I was trying to simply rebase in the changes from today. I am failing at today. I am trying to fix whatever just happened. ::facepalm::

plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Applications.kt
```diff
@@ -0,0 +1,16 @@
...
+import java.io.File
+import java.io.FileFilter
+
+internal object ApkFileFilter : FileFilter {
+    private const val APK_EXTENSION = "apk"
+
+    override fun accept(file: File) = file.extension.toLowerCase() == APK_EXTENSION
+}
+
+internal object BundleFileFilter : FileFilter {
```
ChristianBecker: What do we do with this `BundleFileFilter`? Or is this included for an upcoming PR?

plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
```diff
@@ -24,8 +25,8 @@ open class PublishApk : PlayPublishPackageBase() {
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:InputFiles
     internal val inputApks by lazy {
-        // TODO: If we take a customizable folder, we can fix #233, #227
-        variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
+        extension.buildInputFolder?.listFiles(ApkFileFilter)?.toList()
+                ?: variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
```
bhurling: We should probably be a bit more specific here. I would think if the user defines a folder through this property, we should fail the build if that folder is empty or does not contain any apk files. Why else would the user specify the folder. We should not fall back to the variant output in that case. In the current implementation we could have the following situation:

- The user defines a build folder
- that folder does not exist (`listFiles(Filter)` returns `null`)
- we would fall back to the variant output, which is probably not what the user expects

So we should rather have the following logic:
- If `buildInputFolder` is not null, and contains apks, upload them
- If `buildInputFolder` is not null, but does not exist or there are no apks, fail the build
- If `buildInputFolder` is null, upload apks from variant output

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
```diff
@@ -111,6 +111,23 @@ class PlayPublishTaskTest {
...
+    @Test
+    void whenPublishing_withOverrideFolder_UploadsOnlySpecifiedFiles() {
+        def project = TestHelper.evaluatableProject()
+        project.play {
+            buildInputFolder new File(TestHelper.FIXTURE_WORKING_DIR, "override_output")
+        }
+        project.evaluate()
+
+        setMockPublisher(project.tasks.publishApkRelease)
+        publishApk(project.tasks.publishApkRelease, "example_1.apk")
```
bhurling: This test looks a little suspicious. Shouldn't that filename be automatically retrieved when calling publish? Seems like we're testing the implementation of `AndroidPublisher.Edits.publishApk(editId: String, apkFile: FileContent)` which is too late. The `inputApks` property is read from the main `@TaskAction`

gtcompscientist: I was able to add a test for the empty/non-existent directory that is set, but I still can't figure out how to trigger the `upload` action for this test.

plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
```diff
@@ -24,8 +25,13 @@ open class PublishApk : PlayPublishPackageBase() {
...
     internal val inputApks by lazy {
-        // TODO: If we take a customizable folder, we can fix #233, #227
-        variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
+        if (extension.buildInputFolder == null) {
+            variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
+        } else {
+            extension.buildInputFolder!!.listFiles(ApkFileFilter).toList().apply {
+                require(isNotEmpty(), { "Build input folder provided, but no APK files found" })
+            }
+        }
```
SUPERCILEX: We should never intentionally throw exceptions in tasks IMO. You know what they say, exceptions are for exceptional conditions. 😆 Anyway, I have it set up so Gradle will show a `NO-SOURCE` thing if the folder is empty which is the established convention. So your code should look something like this:
```kt
val override = extension.buildInputFolder
@Suppress("IfThenToElvis") // listFiles is nullable so the elvis would be wrong
if (override == null) {
    variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
} else {
    override.listFiles(ApkFileFilter).orEmpty().toList()
}
```

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
```diff
@@ -111,6 +112,34 @@ class PlayPublishTaskTest {
         doReturn(editCommitMock).when(editsMock).commit(anyString(), anyString())
     }
 
+    @Test
+    void whenPublishing_withOverrideFolder_UploadsOnlySpecifiedFiles() {
```
SUPERCILEX: Yeah, Gradle doesn't support testing incremental inputs very well. I personally don't see the point of testing a very simple piece of code so I say we remove the tests so you can focus on the live testing stuff which will be way more useful.

plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
bhurling: Looks interesting. I agree we should show `NO-SOURCE` if the input folder is empty. However, there's still a case where that input folder doesn't even exist. listFiles() would be null and we'd throw a null pointer exception. That shouldn't happen. For that case we should either

- throw a custom exception with a proper message
- log warning and show `NO-SOURCE`

Nimrodda:
Regarding the custom input dir configuration introduced in this PR. How would that work in a multi variant project?

@SUPERCILEX 
>Don't depend on the assemble task if using custom input dir

I think it would be better not depend on assemble task at all regardless of custom input dir. IMO, assemble can be specified on demand by the developer prior to the publish task.

SUPERCILEX:
@Nimrodda Yeah, we didn't think of that either. Gonna close this until we can put more thought into it.

alexfdz:
@SUPERCILEX still aiming for 2.0? Any other work around to provide gradle-play-publisher existing APKs?

SUPERCILEX:
@alexfdz Probably not. @bhurling is saturated and we need to think more carefully about the API. In the meantime, you can always fork this repo and customize the `PublishApk` task to your liking.

PR #299, opened 2018-06-04
Fix bugs found in production

SUPERCILEX:
@bhurling or @ChristianBecker This should be merged normally since each commit addresses a different bug. (I was too lazy to make a gazillion PRs.)


bhurling:
LGTM. Waiting for confirmation that null is acceptable non-rollout tracks

SUPERCILEX:
@bhurling I tried it myself with the alpha and internal tracks and it works fine.

PR #301, opened 2018-06-04
Add version code conflict resolution strategies

### LINKED ISSUE #168: Upload only if Local App version is higher than app version in Play Store

SUPERCILEX:
Fixes #168


plugin/src/main/kotlin/com/github/triplet/gradle/play/ProcessPackageMetadataTask.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.gradle.api.tasks.TaskAction
+
+open class ProcessPackageMetadataTask : PlayPublishTaskBase() {
+    @TaskAction
+    fun processVersionCodes() = read { editId ->
+        val maxVersionCode = bundles().list(variant.applicationId, editId).execute().bundles
+                ?.map { it.versionCode }
+                ?.max() ?: apks().list(variant.applicationId, editId).execute().apks
+                ?.map { it.versionCode }
+                ?.max() ?: 1
```
gtcompscientist: Shouldn't the unfound `maxVersionCode` be either `0` or `null`? In practice, it isn't going to be unfound for anything that we can upload, though.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ProcessPackageMetadataTask.kt
```diff
@@ -0,0 +1,30 @@
...
+
+        when (extension._resolutionStrategy) {
+            ResolutionStrategy.AUTO -> variant.outputs.filterIsInstance<ApkVariantOutput>().forEach {
+                val newCode = maxVersionCode + 1
+                it.versionCodeOverride = newCode
+                it.versionNameOverride = "${variant.versionName}.$newCode"
```
gtcompscientist: Is this something that should be configurable? I know not everybody uses this naming convention.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ProcessPackageMetadataTask.kt
SUPERCILEX: I'm using 1 because if the API lets us upload the first APK without using the console, 1 is the default. But yeah, mostly it's just to make the Kotlin type system happy.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -97,13 +97,24 @@ class PlayPublisherPlugin : Plugin<Project> {
...
 
             if (variant.isSigningReady) {
+                val processPackageMetadata = project.newTask<ProcessPackageMetadataTask>(
+                        "processPackageMetadata$variantName",
+                        "Processes packaging metadata for $variantName.",
+                        null
+                ) {
+                    init()
+
+                    variant.checkManifest.dependsOn(this)
```
bhurling: I am highly concerned about this task dependency. Since the `check[Variant]Manifest` is run very early in the build chain, we would introduce a call to the play store API even though in most cases we don't have to. E.g. check this tree:

```
--- :app:generateReleaseBuildConfig
     \--- :app:checkReleaseManifest
          \--- :app:preReleaseBuild
               \--- :app:preBuild
```

generating the build config is usually even triggered by Android Studio after syncing the gradle scripts.

Could we find a way to run this task only when absolutely necessary?

plugin/src/main/kotlin/com/github/triplet/gradle/play/ProcessPackageMetadataTask.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import com.android.build.gradle.api.ApkVariantOutput
+import com.github.triplet.gradle.play.internal.PlayPublishTaskBase
+import com.github.triplet.gradle.play.internal.ResolutionStrategy
+import org.gradle.api.tasks.TaskAction
+
+open class ProcessPackageMetadataTask : PlayPublishTaskBase() {
+    @TaskAction
+    fun processVersionCodes() = read { editId ->
+        val maxVersionCode = bundles().list(variant.applicationId, editId).execute().bundles
```
bhurling: Any reason why we are not using the tracks API here? We kind of imply that once users switched to bundles, they never go back to APKs (as we would resolve to the highest version code for bundles no matter if there's a higher apk version). 

Plus, I believe that the `bundles()/apks()` API would also list non-released versions from the artifact archive. Is that the desired behavior?

SUPERCILEX: You're right, I'm testing the tracks API right now and I'll push if it works.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
SUPERCILEX: Yeah, I know. Sadly, there's no way around this because once the manifest task is run, our version code overrides are ignored. My previous implementation was even worse, it queried stuff at configuration time! 🤦‍♂️ Here's why this isn't such a big deal: our tasks are only applicable for release builds which means normal development won't be affected.

Either way, I think your fear is well founded and I'll short circuit the query for `IGNORED` so the default isn't affected.

plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/ResolutionStrategy.kt
```diff
@@ -0,0 +1,7 @@
+package com.github.triplet.gradle.play.internal
+
+enum class ResolutionStrategy(val publishedName: String) {
```
ChristianBecker: `internal`?

SUPERCILEX: Whoops, good catch.

bhurling:
I don't think this PRs serves the purpose we were initially going after in #168. It sees like with these changes we would just fail a little faster, while the `IGNORE` strategy would fail during upload. In case of batch uploads for multi-flavor projects, this strategy doesn't help much. I think we can have a much easier solution, where we don't need a new task: 

We would just have two resolution strategies: IGNORE and FAIL (default). IGNORE would just catch specific errors that arise during apk upload, namely the `apkUpgradeVersionConflict` error, whereas the FAIL option would just re-throw.

SUPERCILEX:
@bhurling I really like your `IGNORE` idea, but I strongly disagree about the purpose of this PR. Rereading that issue, it does seem more focused on just ignoring incorrect version codes, but my main focus is #248. It creates a really nice flow: push to a CI server and automatically deploy the next version.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -40,6 +41,27 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
      */
     var errorOnSizeLimit = true
 
+    internal var _resolutionStrategy = ResolutionStrategy.IGNORE
```
bhurling: We should default to FAIL here as that would keep the original behavior of 1.x. And I think it makes sense, ignoring upload errors should be a conscious decision

plugin/src/main/kotlin/com/github/triplet/gradle/play/ProcessPackageMetadataTask.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import com.android.build.gradle.api.ApkVariantOutput
+import com.github.triplet.gradle.play.internal.PlayPublishTaskBase
+import com.github.triplet.gradle.play.internal.ResolutionStrategy
+import org.gradle.api.tasks.TaskAction
+
+open class ProcessPackageMetadataTask : PlayPublishTaskBase() {
+    @TaskAction
+    fun process() {
+        if (extension._resolutionStrategy != ResolutionStrategy.IGNORE) processVersionCodes()
```
bhurling: Let's only process version codes if strategy is AUTO. The FAIL case in `processVersionCodes()` is a nice shortcut but I think it's not worth the extra server roundtrip just to fail a little faster. Remember, we would introduce this check (and fail the build) even if the developer did not even intent to publish the apk

SUPERCILEX:
@bhurling I agree with you on both accounts and 30b5a2d should address those concerns. 👍

PR #304, opened 2018-06-05
Add support for Gradle cache 

### LINKED ISSUE #267: Implement gradle caching
This would remove the need for the `uploadImages` flag and a bunch of other pointless uploading since Gradle would tell us when a file has changed.

SUPERCILEX:
@bhurling This one should also be merged normally because it fixes a somewhat unrelated bug.

Fixes #267

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
@@ -1,12 +1,20 @@
 package com.github.triplet.gradle.play
 
 import com.github.triplet.gradle.play.internal.AccountConfig
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import java.io.File
 
 data class PlayAccountConfigExtension @JvmOverloads constructor(
-        internal val name: String = "", // Needed for Gradle
+        @get:Internal internal val name: String = "", // Needed for Gradle
 
+        @get:PathSensitive(PathSensitivity.RELATIVE) @get:InputFile @get:Optional
         override var jsonFile: File? = null,
+        @get:PathSensitive(PathSensitivity.RELATIVE) @get:InputFile @get:Optional
         override var pk12File: File? = null,
+        @get:PathSensitive(PathSensitivity.RELATIVE) @get:InputFile @get:Optional
         override var serviceAccountEmail: String? = null
 ) : AccountConfig
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
@@ -4,13 +4,17 @@ import com.github.triplet.gradle.play.internal.AccountConfig
 import com.github.triplet.gradle.play.internal.ReleaseStatus
 import com.github.triplet.gradle.play.internal.ResolutionStrategy
 import com.github.triplet.gradle.play.internal.TrackType
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Internal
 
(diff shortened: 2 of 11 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
```diff
@@ -1,12 +1,20 @@
...
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import java.io.File
 
 data class PlayAccountConfigExtension @JvmOverloads constructor(
-        internal val name: String = "", // Needed for Gradle
+        @get:Internal internal val name: String = "", // Needed for Gradle
```
ChristianBecker: The comment can be shifted as `value` into the annotation:
```
@get:Internal("Needed for Gradle")
```

Not sure what we should prefer, though…

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -3,13 +3,17 @@ package com.github.triplet.gradle.play
 import com.github.triplet.gradle.play.internal.AccountConfig
 import com.github.triplet.gradle.play.internal.ReleaseStatus
 import com.github.triplet.gradle.play.internal.TrackType
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Internal
 
 open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension() {
+    @get:Internal
```
ChristianBecker: Explain that this is a backing property via annotation value? (see above)

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
SUPERCILEX: Let's leave it as is because we can't remove the property at all. If it were just a question of visibility, then we would move the comment to the annotation.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
SUPERCILEX: Done

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
ChristianBecker: SGTM

ChristianBecker:
@SUPERCILEX : Would it be possible to add some reasonable [tests for the caching stuff](https://docs.gradle.org/current/userguide/build_cache.html#sec:task_output_caching_inputs)?

SUPERCILEX:
@ChristianBecker what would we test? The generate task already has tests for inputs and outputs. As for the other two, they only have outputs so Gradle can consider them up-to-date if the inputs didn't change.

gtcompscientist:
@ChristianBecker and @SUPERCILEX I would say that if there's anything that could be tested, I can investigate it on my live-testing branch.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -23,29 +27,28 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
...
      * Default is 10% == 0.1.
      */
+    @get:Input
     var userFraction = 0.1
-
-    /**
-     * Choose whether or not to upload images when publishing the Play Store listing. Disabled by
-     * default for performance reasons.
-     */
-    var uploadImages = false
```
bhurling: Let's create a doc issue for this change because we need to mention this in migration hints. People have to remove this property from their gradle files

SUPERCILEX: Done: #315

PR #305, opened 2018-06-05
Split AppDetail out of ListingDetail

ChristianBecker:
They were never used together in the first place.

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Listings.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Listings.kt
@@ -3,12 +3,15 @@ package com.github.triplet.gradle.play.internal
 import java.io.File
 import java.io.FileFilter
 
+// region '419' is a special case in the Play Store that represents latin america
+// 'fil' is a special case in the Play Store that represents Filipino
+private val localeRegex = Regex("^(fil|[a-z]{2}(-([A-Z]{2}|419))?)\\z")
+
 internal object LocaleFileFilter : FileFilter {
-    // region '419' is a special case in the Play Store that represents latin america
-    // 'fil' is a special case in the Play Store that represents Filipino
-    override fun accept(file: File) =
-            file.name.matches(Regex("^(fil|[a-z]{2}(-([A-Z]{2}|419))?)\\z"))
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Listings.kt
```diff
@@ -3,13 +3,16 @@ package com.github.triplet.gradle.play.internal
 import java.io.File
 import java.io.FileFilter
 
+// region '419' is a special case in the Play Store that represents latin america
+// 'fil' is a special case in the Play Store that represents Filipino
+private val localeRegex = Regex("^(fil|[a-z]{2}(-([A-Z]{2}|419))?)\\z")
```
SUPERCILEX: Can we move this and the `imageExtensions` inside their respective filter objects? Makes it clearer what they're used for.

plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Listings.kt
```diff
@@ -20,23 +23,28 @@ internal enum class ListingDetail(val fileName: String, val maxLength: Int = Int
...
-    DEFAULT_LANGUAGE("defaultLanguage");
+    DEFAULT_LANGUAGE("defaultLanguage")
 }
 
-internal enum class ImageType(val fileName: String, val maxNum: Int = 8) {
-    ICON("icon", 1),
-    FEATURE_GRAPHIC("featureGraphic", 1),
-    PROMO_GRAPHIC("promoGraphic", 1),
+// Min length for any side: 320px. Max length for any side: 3840px.
+internal data class ImageSize(val minWidth: Int, val minHeight: Int, val maxWidth: Int = minWidth, val maxHeight: Int = minHeight)
```
SUPERCILEX: This is and the enum declaration below are too long. Should be like this:
```kt
internal data class ImageSize(
        val minWidth: Int,
        val minHeight: Int,
        val maxWidth: Int = minWidth,
        val maxHeight: Int = minHeight
)
```

Also, can we move the data class below the enum? Standard practice is to go `fields -> methods -> objects -> enums -> classes`.

ChristianBecker:
@SUPERCILEX : PTAL

PR #308, opened 2018-06-06
Add incremental compilation support

SUPERCILEX:
Built on top of #304 which makes it so that as long as you don't change anything, nothing new will be uploaded. This PR goes a step further and incrementally uploads changes.


plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
```diff
@@ -107,57 +107,6 @@ class PlayPublishTaskTest {
...
-        // Attach the mock
-        setMockPublisher(project.tasks.publishApkPaidRelease)
-
-        // finally run the task we want to check
-        project.tasks.publishApkPaidRelease.publishApks()
-
-        // verify that we init the connection with the correct application id
-        verify(editsMock).insert('com.example.publisher.paid.release', null)
-    }
-
```
bhurling: Why did we have to remove those tests? Those tests shouldn't be broken because they make sure we upload to the correct application ID

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
```diff
@@ -273,68 +222,35 @@ class PlayPublishTaskTest {
...
-        // Attach the mock
-        setMockPublisher(project.tasks.publishApkPaidRelease)
-
-        // finally run the task we want to check
-        project.tasks.publishApkPaidRelease.publishApks()
-
-        // verify that we init the connection with the correct application id
-        verify(editsMock).insert('com.example.publisher.paid.release', null)
-    }
```
bhurling: Same here. Did this test break?

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
SUPERCILEX: Yeah, they broke because there isn't an easy way to call the @TaskAction method with fake inputs. In the end, I decided to remove the tests because they're literally just checking one line:
https://github.com/Triple-T/gradle-play-publisher/blob/239d3aae4d86a66fd0f9b400589d8090e04347d1/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/PlayPublishTaskBase.kt#L52

And it's not even specific to the APK task. If you want, we could add a test for that but I don't really see the point in check a simple line of code. 🤷‍♂️

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
SUPERCILEX: Yup, see [above](https://github.com/Triple-T/gradle-play-publisher/pull/308#discussion_r193976041).

plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublishTaskTest.groovy
bhurling: Those tests were actually quite useful at the time. In #54, #55 we found out that reading the application id from the variant at runtime makes sure we pick up any changes that happen after evaluation (which we didn't support before).

Completely removing them is a bit scary. Plus, we definitely need a way to call task actions in tests even with the new input parameters. We should at least keep one of them or a simplified version (preferably the one about changing the application id after evaluate). That would also serve as a template of how to mock task inputs

bhurling: Found an easy way to bring back that test. I'll add a commit if you don't mind

PR #319, opened 2018-06-08
Add support for Android App Bundles

### LINKED ISSUE #262: New Android App Bundle
Hi,
Do you plan to integrate the new "Android App Bundle" feature ?
[https://android-developers.googleblog.com/2018/05/io-2018-everything-new-in-google-play.html](https://android-developers.googleblog.com/2018/05/io-2018-everything-new-in-google-play.html)
[https://www.youtube.com/watch?v=bViNOUeFuiQ](https://www.youtube.com/watch?v=bViNOUeFuiQ)

the new command is : 'gradlew app:bundle'
[https://youtu.be/bViNOUeFuiQ?t=14m10s](https://youtu.be/bViNOUeFuiQ?t=14m10s)

Thanks for all 
Thomas :)

SUPERCILEX:
Built on top of #347

Fixes #262


plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishBundle.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.gradle.api.tasks.incremental.IncrementalTaskInputs
+import java.io.File
+
+open class PublishBundle : PlayPublishPackageBase() {
+    @Suppress("MemberVisibilityCanBePrivate", "unused") // Used by Gradle
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    @get:InputFile
+    val bundle by lazy {
+        // TODO: If we take a customizable folder, we can fix #233, #227
+        File(project.buildDir, "outputs/bundle/${variant.name}/bundle.aab")
```
knordbo: I just built this locally and ran it and my .aab happened to be app.aab. Using AGP 3.2.0-beta01. Probably should not hardcode this one?

SUPERCILEX: Yeah, I'm running into the same issue with `beta02` in some production apps. Not in my test project though, still investigating.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -10,8 +10,7 @@ import org.gradle.api.tasks.Internal
 
 open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension() {
     /**
-     * Choose the default packaging method. Either App Bundles or APKs. Affects tasks like
-     * `publish[Variant]`.
+     * Choose the default packaging method. Either App Bundles or APKs. Affects tasks like publish`.
```
ChristianBecker: Missing backtick?

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -175,7 +175,9 @@ class PlayPublisherPlugin : Plugin<Project> {
                 project.tasks.findByName(
                         (variant as InstallableVariantImpl).variantData.getTaskName("bundle", ""))
                         ?.let { dependsOn(it) }
-                        ?: logger.warn("Bundle task not found. Publishing App Bundles may not work.")
+                        ?: logger.warn("Bundle task not found, make sure to use " +
+                                               "'com.android.tools.build:gradle:3.2.+'. " +
```
ChristianBecker: Any chance we can change this to not suggest wildcard versioning?

SUPERCILEX: Yeah, I was trying to say "use 3.2+" but it looks like I'm telling devs to add that dependency. 👻

knordbo:
One problem I'm seeing locally is that `releaseNotesDir` in `PlayPublishPackageBase` tries to resolve, even if I'm just doing a publish task. That crashes the build. Not sure if that is part of this PR to fix though.

SUPERCILEX:
@knordbo I'm not sure how update-to-date this PR is relative to master. We fixed an issue that caused a crash if there was no `release-notes` directory a few days ago. 🤷‍♂️ If that's not the issue, feel free to share.

knordbo:
@SUPERCILEX this branch is synced with master. I think the two lines here which originally had the fix was accidentally reverted: https://github.com/Triple-T/gradle-play-publisher/commit/525b13203462e3417c68d5e3a8d158c5cf2d073d#diff-50be7e56abb0dd9419edb1f897d71a58 the `orNull()` does the trick since that directory is not guaranteed to be there, even if you created it at that point in time it will still be wiped by clean later, so safer to keep it nullable AFAIK.

SUPERCILEX:
@knordbo would you mind filing a bug report with your exact issue so we don't clutter this PR with unrelated stuff? Those changes were intentional.

PR #320, opened 2018-06-09
Remove one Ignore'd test

ChristianBecker:
The removed test was Ignore'd for a while now. The test itself was unrelated to our plugin.


plugin/src/test/groovy/com/github/triplet/gradle/play/PlayPublisherPluginTest.groovy
```diff
@@ -296,8 +296,8 @@ class PlayPublisherPluginTest {
 
         project.evaluate()
 
-        project.extensions.play.serviceAccountEmail == 'service-account@test.com'
-        project.extensions.play.pk12File = new File('key.p12')
+        assertEquals('service-account@test.com', project.extensions.play.serviceAccountEmail)
```
ChristianBecker: This was a bug. It hasn't tested anything.

PR #322, opened 2018-06-10
Add support for in-app products

### LINKED ISSUE #181: Add support for in-app products

SUPERCILEX:
Fixes #181


gtcompscientist:
Since this is built on top of another PR, I can't tell easily. Did you also add bootstrapping those?

SUPERCILEX:
Yup:
https://github.com/Triple-T/gradle-play-publisher/blob/386f09bc9477f5d4cf438807d65b1aaad98d621a/plugin/src/main/kotlin/com/github/triplet/gradle/play/BootstrapTask.kt#L115-L120

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -120,6 +127,17 @@ class PlayPublisherPlugin : Plugin<Project> {
                 doFirst { logger.warn("$name is deprecated, use ${publishListingTask.get().name} instead") }
             }
 
+            val publishProductsTask = project.newTask<PublishProducts>(
+                    "publish${variantName}Products",
+                    "Uploads all Play Store in-app products for $variantName."
+            ) {
+                init()
+                productsDir = File(playResourcesTask.get().resDir, PRODUCTS_PATH)
```
bhurling: Let's move this into a lazy block inside the task as in #354

SUPERCILEX: 👍

PR #323, opened 2018-06-10
Merge resource translations given default locale

### LINKED ISSUE #107: Improve missing language support
I am managing a growing number of application variants (and this plugin is a godsend, thanks!) however one of the differences between them is the languages that they support. For example there are variants with en-GB or en-US depending on the location, and others which support local languages like Catalan. I would like to define a `whatsnew` file for every possible language in the `src/main/play` folder but if I do that then I get a `apkRecentChangesAddedForMissingLanguage` error for any language not defined for a particular app. 

```
> com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
  {
    "code" : 403,
    "errors" : [ {
      "domain" : "androidpublisher",
      "message" : "Tried to set recent changes text for APK version 179 for language en-GB. Language is not associated with the app.",
      "reason" : "apkRecentChangesAddedForMissingLanguage"
    } ],
    "message" : "Tried to set recent changes text for APK version 179 for language en-GB. Language is not associated with the app."
  }
```

Is there a workaround for this error? Could it be skipped for apps which don't support a given language? I would like to avoid having to define specific language folders for every variant because that would mean having to maintain `whatsnew` files for each one, which is exactly what I'm trying to avoid.

SUPERCILEX:
Built on top of #311

Fixes #107


plugin/src/test/groovy/com/github/triplet/gradle/play/GenerateResourcesTest.groovy
```diff
@@ -367,4 +367,46 @@ class GenerateResourcesTest {
...
+                'src/staging/play/listings/en-US/shortdescription').text
+
+        project.android {
+            flavorDimensions 'pricing', 'server'
+
+            productFlavors {
+                free { dimension 'server' }
+                paid { dimension 'pricing' }
+                staging { dimension 'server' }
+                prod { dimension 'pricing' }
```
bhurling: adding prod flavor to pricing dimension. Is that on purpose to be able to run the desired check?

plugin/src/test/groovy/com/github/triplet/gradle/play/GenerateResourcesTest.groovy
```diff
@@ -367,4 +367,46 @@ class GenerateResourcesTest {
...
+            }
+
+            buildTypes {
+                dogfood.initWith(buildTypes.release)
```
bhurling: not using this in the test. Can be removed

plugin/src/main/kotlin/com/github/triplet/gradle/play/GenerateResourcesTask.kt
```diff
@@ -30,6 +35,12 @@ open class GenerateResourcesTask : DefaultTask() {
     private val resSrcDirs: List<File> by lazy {
         variant.sourceSets.map { project.file("src/${it.name}/$PLAY_PATH") }
     }
+    private val defaultLocale by lazy {
+        resSrcDirs.mapNotNull {
+            File(it, AppDetail.DEFAULT_LANGUAGE.fileName).orNull()
+                    ?.readText()?.normalized().nullOrFull()
+        }.lastOrNull()
```
bhurling: Can you explain why `lastOrNull()` is the right option here?

SUPERCILEX: The `resSrcDirs` are sorted from lowest to highest priority. So taking the last element means getting the highest priority override—it's basically a more efficient way of reversing the list and searching for the most specific override. FYI, the `resSrcDirs` looks something like this: `main, flavor, buildType, flavorBuildType`.

plugin/src/test/groovy/com/github/triplet/gradle/play/GenerateResourcesTest.groovy
SUPERCILEX: Yeah, I had to move the dimensions and flavors around to get the overrides (or lack thereof) I wanted.

PR #331, opened 2018-06-14
Update AndroidPublisher to v3-rev12-1.23.0

ChristianBecker:
Unfortunately, there are no release notes available. But the build is green, so we're good to go!

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 
 dependencies {
     implementation("com.android.tools.build:gradle:3.0.1")
-    implementation("com.google.apis:google-api-services-androidpublisher:v3-rev6-1.23.0") {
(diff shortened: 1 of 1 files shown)


plugin/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
 
 dependencies {
     implementation("com.android.tools.build:gradle:3.0.1")
-    implementation("com.google.apis:google-api-services-androidpublisher:v3-rev6-1.23.0") {
+    implementation("com.google.apis:google-api-services-androidpublisher:v3-rev11-1.23.0") {
```
SUPERCILEX: They've already pumped out `rev12` 😆. I say we update it this time and then just stop for a long while. Otherwise, we'll constantly be updating it.

ChristianBecker: Yeah, I updated it and we should probably delay the constant update to the releases..?

SUPERCILEX: SGTM

PR #335, opened 2018-06-15
Modify release

gtcompscientist:
Implementation of the modification tasks. Still need to implement tests, but wanted to get feedback on naming conventions, and the basic setup.

After the changes to the different tasks of late, I think that this is the simplest method for this task.


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -23,6 +23,21 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
...
         }
 
+    @get:Internal("Backing property for CLI option")
+    internal var _fromTrack = TrackType.INTERNAL
+    /**
+     * Specify the track for your app that will be modified. May be one of internal, alpha, beta, rollout,
+     * or production. Default is internal.
+     */
+    @get:Input
+    internal var fromTrack
```
gtcompscientist: I think that this should be entirely `internal` and the modification should require the CLI option.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ModifyReleaseTask.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.github.triplet.gradle.play.internal.TrackType
+import org.gradle.api.tasks.TaskAction
+
+open class ModifyReleaseTask : PlayPublishPackageBase() {
+    init {
+        // Always out-of-date since we don't know what's changed on the network
+        outputs.upToDateWhen { false }
+    }
+
+    private var startCodes: List<Long>? = null
```
SUPERCILEX: Let's make this a local variable so we don't have to worry about mutability or methods being called in a specific order.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ModifyReleaseTask.kt
```diff
@@ -0,0 +1,43 @@
...
+    @TaskAction
+    fun modify() {
+        progressLogger.start("Modifies release tracks for variant ${variant.name}", null)
+
+        requireNotNull(extension._fromTrack) {
+            progressLogger.completed("No from-track provided to modify", true)
+        }
+        downloadVersionCodes(extension._fromTrack!!)
+
+        startCodes.let {
```
SUPERCILEX: Making it a local variable will also let us get rid of the unnecessary `let`.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ModifyReleaseTask.kt
```diff
@@ -0,0 +1,43 @@
...
+            if (it == null || it.isEmpty()) {
+                progressLogger.completed("No versions published on track '${extension.track}' for ${variant.name}", false)
+            } else {
+                write { editId: String ->
```
SUPERCILEX: This write should be pulled up and the `downloadVersionCodes` should use that edit instead of creating a read and write.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ModifyReleaseTask.kt
```diff
@@ -0,0 +1,43 @@
...
+        progressLogger.start("Modifies release tracks for variant ${variant.name}", null)
+
+        requireNotNull(extension._fromTrack) {
+            progressLogger.completed("No from-track provided to modify", true)
+        }
+        downloadVersionCodes(extension._fromTrack!!)
+
+        startCodes.let {
+            if (it == null || it.isEmpty()) {
+                progressLogger.completed("No versions published on track '${extension.track}' for ${variant.name}", false)
```
SUPERCILEX: The Gradle API is kinda bad. Your final status won't be visible because the task will finish immediately afterwards. Basically, what I'm saying is that there should only be one `completed` call without a status at the end of the method.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
SUPERCILEX: Why do we need a `fromTrack` at all? It makes sense to me that you'd want to shift the track above your target to the left. So say the target is `beta`, we'd move `alpha` into beta, or if it isn't available, `internal`. Another example: our target is `production` so we take `beta` and move it over (or `alpha` if it isn't available and so on). That way we mirror behavior from the Play Console too.

gtcompscientist: You don't have to use all of the tracks, and you can skip tracks, and have releases in multiple tracks. I may want to promote everything from Alpha to Production, and Internal to Beta (Which is SOP on the Play Console), versus having to march through each of those steps with subsequent commands.

plugin/src/main/kotlin/com/github/triplet/gradle/play/ModifyReleaseTask.kt
gtcompscientist: Interesting, seems like a poor design.

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
gtcompscientist: For the sake of clarity, I will say that I initially considering using `toTrack` instead of `fromTrack`, because the functionality on the Play Store is: 
"Pick an exiting TrackRelease"->"Choose what Track to promote it to"
Which would _normally_ mean a user would be using their default from their normal gradle script as the start track, but that would mean modifying the `updateTracks` method, and it seemed better to have this specific override.

SUPERCILEX: Hmmm, sounds fair

gtcompscientist:
Being able to promote from the command-line/CI (via tags) means I can automate a lot of the process, and remove myself from that process more. That means that I can give somebody permissions through a backend service we own and let them push a single button to modify one, or more apps. I understand that somebody could be given permission to the Google Play Console, but that gets REALLY ugly when you have as many apps as I have, and they may or may not be within the same org.

For your second point, are you saying that the CLI option wouldn't work for the bulk option?

For variants not having a current release on the `fromTrack`, I would say that it should be skipped with a warning.

Regarding your last note: I could be convinced to call it `promote`, but it's not just promoting. It is there to be used for:

- **modifying** release notes
- **modifying** user rollout fraction
- promoting tracks

gtcompscientist:
@bhurling This should now be ready for review and merging. Deep testing will need to be part of the `live-testing` effort. In actuality, they'll be crucial to the process.

SUPERCILEX:
@gtcompscientist can we get through the build input stuff first? #296 Side note: this PR should probably come after #353 so you can integrate it.

Also, I disagree about not thoroughly testing this—everything that goes into master should ideally be guaranteed to work (or as close as possible). And then I'd like to think about the API a tad bit more (though that could be done later).

SUPERCILEX:
Closing in favor of #389

PR #347, opened 2018-06-21
Release notes fixes and other annotation improvements

SUPERCILEX:
- Make release notes dir optional (aka don't crash if it isn't there)
   - Remove pointless network call updating tracks
- Update out-of-date credentials annotations

Closes #358

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/PlayPublishPackageBase.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/PlayPublishPackageBase.kt
@@ -18,12 +18,12 @@ abstract class PlayPublishPackageBase : PlayPublishTaskBase() {
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:Optional
     @get:InputDirectory
-    internal val releaseNotesDir by lazy { File(resDir, RELEASE_NOTES_PATH).orNull() }
+    internal val releaseNotesDir by lazy { File(resDir, RELEASE_NOTES_PATH) }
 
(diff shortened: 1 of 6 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/PlayPublishPackageBase.kt
```diff
@@ -41,8 +44,10 @@ abstract class PlayPublishPackageBase : PlayPublishTaskBase() {
             versionCodes = versions
         }
 
-        track.releases = listOf(trackRelease)
-
+        val track = Track().apply {
```
bhurling: We're now always creating a new Track object instead of trying to update an existing one. Is that safe to do?

SUPERCILEX: Yeah: the track only has two fields and we overwrite both of them... So there was no point to what we were doing before. 😁

SUPERCILEX: Well for the `Track#track` field, we just found one that matched the `extension.track` which is the same as just setting it directly if you're going to overwrite the releases.

bhurling: Never mind, there's no other state than the list of releases and we're setting our own list anyway.

SUPERCILEX: Yup 👍

PR #350, opened 2018-06-22
Add contributing guidelines

SUPERCILEX:


bhurling:
Shouldn't we have this file in the root directory? It seems like a file that is independent of code hosting. Anything we gain by "hiding" it in the .github folder?

SUPERCILEX:
You're right, moved 👍

PR #354, opened 2018-06-22
Minimize mutability

SUPERCILEX:
- Make properties protected where possible (lowest permissions Gradle will read annotations from)
- Make properties val where possible.


plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
```diff
@@ -23,14 +23,14 @@ open class PublishApk : PlayPublishPackageBase() {
...
-    internal val inputApks by lazy {
+    protected val inputApks by lazy {
         // TODO: If we take a customizable folder, we can fix #233, #227
         variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
     }
     @Suppress("MemberVisibilityCanBePrivate", "unused") // Used by Gradle
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:OutputDirectory
-    internal val outputDir by lazy { File(project.buildDir, "${variant.playPath}/apks") }
+    protected val outputDir by lazy { File(project.buildDir, "${variant.playPath}/apks") }
```
bhurling: Why protected? We're not extending `PublishApk`. I wonder why our task classes have to be `open` in the first place

ChristianBecker: The tasks have to be `open` so that Gradle can create proxy classes for our tasks, at least that's my understanding of it.

SUPERCILEX: @ChristianBecker is correct about the `open` stuff. Similarly, the generated getter has to be at least protected for Gradle's annotations to work.

Issue #306, opened 2018-06-05
Remove test dependencies and android gradle plugin from published pom


bhurling:
When publishing the artifact on the 2.x branch we currently include all the test scoped dependencies. That doesn't seem to be good practice as it would force downstream clients to download these dependencies as well. On top of that it could make sense to specify the android plugin as `provided` or `compileOnly` so we would not bring it ourselves but instead rely on the clients to specify their version (which is most certainly the case for any android project anyway)

ChristianBecker:
FYI: With proper `SNAPSHOT` releases (see #307) it's no show-stopper anymore as all `test`-scoped dependencies have a proper `version` now.

PR #356, opened 2018-06-26
Exclude test dependencies from final pom

bhurling:
Addresses first half of #306 so we no longer dump test dependencies into the published pom

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -35,8 +35,8 @@ gradlePlugin {
 }
 
 afterEvaluate {
-    fun MavenPom.removeTestDependencies() {
-        whenConfigured {
+    fun PomFilterContainer.removeTestDependencies() {
+        pom.whenConfigured {
             dependencies.removeIf {
                 // Stolen from JetBrains' own sample at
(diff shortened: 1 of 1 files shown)


plugin/build.gradle.kts
```diff
@@ -31,3 +33,23 @@ gradlePlugin {
...
+        pom.whenConfigured {
+            dependencies.removeIf {
+                // Stolen from JetBrains' own sample at
+                // https://github.com/JetBrains/kotlin/blob/v1.2.50/buildSrc/src/main/kotlin/plugins/PublishedKotlinModule.kt#L86
+                InvokerHelper.getMetaClass(it).getProperty(it, "scope") == "test"
+            }
+        }
+    }
+
+    (tasks["uploadArchives"] as? Upload?)?.let {
```
SUPERCILEX: This should be `tasks.withType<Upload> { ... }`

plugin/build.gradle.kts
```diff
@@ -31,3 +33,23 @@ gradlePlugin {
...
+        (it.repositories["mavenDeployer"] as PomFilterContainer).removeTestDependencies()
+    }
+
+    (tasks["install"] as? Upload?)?.let {
```
SUPERCILEX: Same here

bhurling: Nice, so we can probably combine it into one statement. I'll try

PR #357, opened 2018-06-26
Use Gradle Plugins repository instead of Bintray

### LINKED ISSUE #154: Add to Gradle Plugins Repository
Please add to https://plugins.gradle.org.

### LINKED ISSUE #306: Remove test dependencies and android gradle plugin from published pom

SUPERCILEX:
Fixes #154, fixes #306, fixes #356


bhurling:
With those changes, what would be the way to publish the plugin into your local maven repository? Previously we had `./gradlew install` to do it. That way you can properly test your changes in a real world setup before opening a PR. Removing the `com.vanniktech.maven.publish` plugin would remove that opportunity. Having added #358 I am currently not very confident that our unit tests catch even basic mistakes and I'd rather have an option to apply any intermediate version to a real world app project

PR #358, opened 2018-06-27
fix input annotations for account configs

bhurling:
This fixes a few annotation errors that made it impossible to actually apply the plugin and run its tasks:

- the service account email is not an `@InputFile`
- credentials file has to be optional as it can be set either through account config or in the play block

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
@@ -1,11 +1,7 @@
 package com.github.triplet.gradle.play
 
 import com.github.triplet.gradle.play.internal.AccountConfig
-import org.gradle.api.tasks.InputFile
-import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.Optional
-import org.gradle.api.tasks.PathSensitive
-import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.*
 import java.io.File
 
 data class PlayAccountConfigExtension @JvmOverloads constructor(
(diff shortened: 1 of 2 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayAccountConfigExtension.kt
```diff
@@ -1,18 +1,14 @@
 package com.github.triplet.gradle.play
 
 import com.github.triplet.gradle.play.internal.AccountConfig
-import org.gradle.api.tasks.InputFile
-import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.Optional
-import org.gradle.api.tasks.PathSensitive
-import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.*
```
ChristianBecker: Can we get rid of star imports?

plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/AccountConfig.kt
```diff
@@ -1,10 +1,6 @@
 package com.github.triplet.gradle.play.internal
 
-import org.gradle.api.tasks.InputFile
-import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.Optional
-import org.gradle.api.tasks.PathSensitive
-import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.*
```
ChristianBecker: Same

bhurling:
Oh, sure

bhurling:
Yes, thought the same while working on this one. I actually prefer multiple lines because it pops out and you immediately see if something is missing

ChristianBecker:
@bhurling : This kind of relates to #347, where @SUPERCILEX adjusted the annotations as part of one commit.

PR #364, opened 2018-07-01
Add compatibility tests

bhurling:
Refactoring the compatibility test suite so it tries to apply our plugin (from inside the `build/jar` folder) to our test fixture project. It will try to set up different combinations of AGP and gradle in the client project. For now we check the following combinations:

- AGP 3.0.1, gradle 4.2
- AGP 3.0.1, gradle 4.3
- AGP 3.1.3, gradle 4.4
- AGP 3.2.0-beta02, gradle 4.6
- AGP 3.2.0-beta02, gradle 4.8

We're basically testing some common combinations where the AGP version dictates the minimum gradle version.

We could use a similar approach if we want to support AGP 2.x and gradle 3.x (but I'd recommend to write a separate test for that as it might require a different build.gradle)


plugin/src/test/kotlin/com/github/triplet/gradle/play/CompatibilityTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    @After
+    fun cleanup() {
+        File(testProject.projectDir, "build.gradle").delete()
+    }
+
+    private companion object {
+        @JvmStatic
+        @Parameterized.Parameters(name = "agpVersion: {0}, gradleVersion {1}")
+        fun parameters() = listOf(
+                // agp 3.0.1, requires at least gradle 4.1
```
bhurling: gradle 4.1 gives me a strange warning of `compileSdkVersion is not specified.` that doesn't happen in production. I'll have to investigate a bit more

PR #375, opened 2018-07-03
Support bootstrapping specific features

### LINKED ISSUE #372: Add CLI options to bootstrap task to limit what is downloaded
As we add new features, devs might want to download only the resources for the latest feature. Case in point: I was working on the in-app purchases support and ran the bootstrap task only to have it overwrite all my files and add duplicate images.

With the new options, we could do something like this: `./gradlew bootstrap --products-only`. Or `./gradlew bootstrap --release-notes-only`.

SUPERCILEX:
Fixes #372

~~Note: this PR should probably be merged normally since it moves files around (no squash).~~

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -9,11 +9,7 @@ plugins {
 }
 
 dependencies {
-    // Using compileOnly because every android project applies
-    // its own version of the Android Plugin.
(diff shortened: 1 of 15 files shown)


SUPERCILEX:
This PR now depends on #397

PR #378, opened 2018-07-10
Use Gradle Kotlin DSL

SUPERCILEX:


plugin/build.gradle.kts
```diff
@@ -2,9 +2,9 @@ import org.codehaus.groovy.runtime.InvokerHelper
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmCompile
 
 plugins {
+    `java-gradle-plugin`
+    `kotlin-dsl`
```
bhurling: any reason to use this syntax instead of the `id('xyz')` syntax? Can we have it consistent for all our plugins?

SUPERCILEX: Yeah, the Gradle team recommends it for their plugins because it ensures the correct version is used when we upgrade our Gradle wrapper. So basically, they pick the version for us:
```kt
val PluginDependenciesSpec.`kotlin-dsl`: PluginDependencySpec
    get() = id("org.gradle.kotlin.kotlin-dsl") version "0.17.5"
```

PR #382, opened 2018-07-27
Fix incorrect app bundle file name detection

### LINKED ISSUE #381: App Bundle upload task doesn't work if project sets archivesBaseName property
### Describe the bug

When project sets `archivesBaseName` to custom value different from `name` property `publishReleaseBundle` task fails.

### To Reproduce

1. Set `archivesBaseNameProperty` to custom value.For example:
```
android {
    defaultConfig {
        versionName "1.0.0"
        archivesBaseName = "$archivesBaseName-$versionName"
    }
}
```
2. Run `./gradlew publishReleaseBundle`
3. Task fails with 
```
A problem was found with the configuration of task ':app:publishReleaseBundle'.
> File '/home/alapshin/workspace/android/foobar/app/build/outputs/bundle/release/app.aab' specified for property 'bundle' does not exist.
```
#### Versions

 - Gradle Play Publisher: 2.0.0-SNAPSHOT
 - Gradle Wrapper (if applicable): 4.9
 - Android Gradle Plugin (if applicable): 3.2.0-beta04

### Expected behavior

Task should correctly detect App Bundle file name and upload it.

### Additional context

Related bug in Google's issuetracker: https://issuetracker.google.com/issues/80552039

alapshin:
Fixes #381

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishBundle.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishBundle.kt
@@ -22,7 +22,8 @@ open class PublishBundle : PlayPublishPackageBase() {
     @get:InputFile
     val bundle by lazy {
         // TODO: If we take a customizable folder, we can fix #233, #227
-        File(project.buildDir, "outputs/bundle/${variant.name}/${project.name}.aab")
+        val archivesBaseName = project.properties["archivesBaseName"] as String
+        File(project.buildDir, "outputs/bundle/${variant.name}/${archivesBaseName}.aab")
     }
     @Suppress("MemberVisibilityCanBePrivate", "unused") // Used by Gradle
     @get:PathSensitive(PathSensitivity.RELATIVE)


alapshin:
Sorry, looks like this property is not available on project variable declared in plugin.
However it is  available when running in custom task declared via file. I will have to think of something else.

alapshin:
I had to query this property using its name similar to how Android Gradle plugin does it:
https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/PackageSplitRes.java#114

If there is a better way to do this please feel free to correct me.

alapshin:
> Will this work if you don't set the base name?

Yes it should work.
By default this property will be equal to `project.name`:
https://docs.gradle.org/current/dsl/org.gradle.api.plugins.BasePluginConvention.html#org.gradle.api.plugins.BasePluginConvention:archivesBaseName

This link is about java plugin but as far as I am aware android plugin follows the same convention.

SUPERCILEX:
Cool beans! Either @bhurling or me will have to check that. (PSA: I'm gone until Monday)

SUPERCILEX:
Oh, I completely forgot about this PR. Sorry! ❤️ Will check later today

Issue #373, opened 2018-07-03
Use new Gradle APIs


SUPERCILEX:
This issue is just for keeping track of what we'll want to use when we decide to kill support for old Gradle versions. This won't happen for v2.0.

### 4.7

- [ ] Constructor args for our tasks: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:passing_arguments_to_a_task_constructor
- [ ] Worker API to parallelize image uploading: https://guides.gradle.org/using-the-worker-api/

### 4.9

- [x] Configuration avoidance for all our tasks: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

oliverlockwood:
Hi @SUPERCILEX, does this issue cover the following warnings that I'm currently getting?

```
> Task :my-project:generateReleasePlayResources
Some problems were found with the configuration of task ':my-project:generateReleasePlayResources'. Registering invalid inputs and outputs via TaskInputs and TaskOutputs methods has been deprecated and is scheduled to be removed in Gradle 5.0.
 - File '/Users/ogl/work/my-project/src/main/play' specified for property '$1' does not exist.
 - File '/Users/ogl/work/my-project/src/release/play' specified for property '$2' does not exist.
 - File '/Users/ogl/work/my-project/src/release/play' specified for property '$3' does not exist.
```

Or have I missed some part of configuration?  (Note that with my current configuration, builds are being published by your plugin to the Play Store just fine).

Note that I have configured https://docs.gradle.org/4.5/userguide/command_line_interface.html#sec:command_line_warnings to `all` to make this visible, rather than just a summary.

SUPERCILEX:
@oliverlockwood no, but this will: https://github.com/Triple-T/gradle-play-publisher/pull/386. That was my B. 😊

SUPERCILEX:
Constructor args: meh

Worker API: we can't use it to upload anything because the Publisher API isn't serializable. 😢

PR #388, opened 2018-08-28
Support task configuration avoidance

SUPERCILEX:
See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

Progress on #373


SUPERCILEX:
Oh, and this is built on top of #378

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -117,56 +117,49 @@ class PlayPublisherPlugin : Plugin<Project> {
...
                     "process${variantName}Metadata",
                     "Processes packaging metadata for $variantName.",
                     null
-            ) {
-                init()
-
-                variant.checkManifest.dependsOn(this)
-            }
+            ) { init() }
+            checkManifest.dependsOn(processPackageMetadata)
```
bhurling: Should this be `checkManifest.configure { dependsOn(processPackageMetadata) }`?

SUPERCILEX: `checkManifest` actually comes from the AGP and they haven't migrated to the new API yet. But yeah, once they do, that's what it will look like. 👍

SUPERCILEX:
Some interesting things to keep in mind with this PR:
- since tasks are only configured when needed, we can't add dependencies in the configuration block or it will never run
- for backcompat reasons, we fake a wrapper class to get real methods at runtime or default to eager configuration on old Gradle versions
- it's ok to call `get` and force eager initialization in the configuration block if it depends on the task since the forcibly created task would be created anyway

PR #389, opened 2018-08-29
Support promoting releases

### LINKED ISSUE #104: Change userFraction after publishing apk

### LINKED ISSUE #284: Add `modifyRelease[Flavor]` task

SUPERCILEX:
Fixes #284, fixes #104

This PR supports everything! 😁
- Moving artifacts to a different track (caveat: the API doesn't support demoting releases so this duplicates the tracks)
- Superseding tracks
- Automatically picking the `fromTrack` if unspecified
- Modifying release notes
- Upping rollout fraction
- Changing status

Example usage to promote `internal` to `alpha`:
```sh
$ ./gradlew publishBundle --track internal
$ ./gradlew promoteArtifact --track alpha
```

Note: the task is named `promoteArtifact` to avoid confusion with variant specific tasks. If it were `promoteRelease`, you would get tasks like `promoteReleaseRelease`. Instead, you get something more reasonable like `promoteReleaseArtifact`.

Example usage to move `beta` to `prod` with user fraction assuming artifacts exist in `alpha`/`internal`:
```sh
$ ./gradlew promoteArtifact \
    --track rollout --from-track beta \
    --user-fraction .4 --release-status inProgress
```

And of course, all of the CLI options are available in the DSL.


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherExtension.kt
```diff
@@ -16,33 +17,54 @@ open class PlayPublisherExtension : AccountConfig by PlayAccountConfigExtension(
     @get:Input
     var defaultToAppBundles = false // App Bundles require Google Play App Signing
 
+    @get:Internal("Backing property for public input")
+    internal var _fromTrack: TrackType? = null
+    /**
+     * Specify the track from which to promote a release. That is, the specified track will be
+     * promoted to [track].
+     *
+     * See [track] for valid values.
```
bhurling: Let's add the default value here

SUPERCILEX: Done!

PR #393, opened 2018-09-02
Don't upload universal APK if splits are provided

### LINKED ISSUE #380: Allow not uploading the universal APK
When you split your APK and have enabled universalApk it will fail:

```
splits {
    abi {
        enable isSplitApk
        reset()
        include 'armeabi-v7a', 'arm64-v8a', 'x86'
        universalApk true
    }
}
```

Because the plugin will try to upload universal apk with other spli apk's and Google play will complain with:

```
:app:publishApkRelease FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:publishApkRelease'.
> com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
  {
    "code" : 403,
    "errors" : [ {
      "domain" : "androidpublisher",
      "message" : "Version 100 of this app can not be downloaded by any devices as they will all receive APKs with higher version codes.",
      "reason" : "multiApkShadowedActiveApk"
    } ],
    "message" : "Version 100 of this app can not be downloaded by any devices as they will all receive APKs with higher version codes."
  }
```

because it will build:

    universal-apk with version 100,
    arm-v7 with version 101,
    arm64-apk with version 102,
    etc...

but universal-apk will not contain more abi's than all other apk versions 101,102...etc

Universal apk is usefull for testing but it would be nice to have some switch to disable upload of this universal apk to Google Play.

Source: [this comment](https://github.com/Triple-T/gradle-play-publisher/issues/39#issuecomment-239446519)

SUPERCILEX:
Fixes #380

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
@@ -1,5 +1,6 @@
 package com.github.triplet.gradle.play.tasks
 
+import com.android.build.VariantOutput.OutputType
 import com.android.build.gradle.api.ApkVariantOutput
 import com.github.triplet.gradle.play.internal.PlayPublishPackageBase
 import com.github.triplet.gradle.play.internal.playPath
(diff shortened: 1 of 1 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishApk.kt
```diff
@@ -24,7 +25,9 @@ open class PublishApk : PlayPublishPackageBase() {
     @get:InputFiles
     internal val inputApks by lazy {
         // TODO: If we take a customizable folder, we can fix #233, #227
-        variant.outputs.filterIsInstance<ApkVariantOutput>().map { it.outputFile }
+        variant.outputs.filterIsInstance<ApkVariantOutput>().filter {
+            OutputType.valueOf(it.outputType) == OutputType.MAIN || it.filters.isNotEmpty()
```
SUPERCILEX: For context, the other output types are splits in which case they have filters determining which type of split they are. So if the output type is a split but doesn't have filters, it means it's got everything aka a universal APK and we ignore it.

PR #404, opened 2018-09-20
Fix deprecations with AGP 3.3

SUPERCILEX:
We have the `CompatibilityTest` to make sure things work on older versions.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -9,7 +9,7 @@ plugins {
 }
 
 dependencies {
-    compileOnly("com.android.tools.build:gradle:3.3.0-alpha11")
+    compileOnly("com.android.tools.build:gradle:3.3.0-alpha12")
 
     implementation("com.google.apis:google-api-services-androidpublisher:v3-rev12-1.23.0") {
         exclude("com.google.guava", "guava-jdk5") // Remove when upgrading to AGP 3.1+
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -9,11 +9,7 @@ plugins {
 }
 
 dependencies {
-    // Using compileOnly because every android project applies
-    // its own version of the Android Plugin.
-    // We're staying on 3.0.1 so we don't accidentally use APIs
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -141,7 +141,12 @@ class PlayPublisherPlugin : Plugin<Project> {
                     "Processes packaging metadata for $variantName.",
                     null
             ) { init() }
-            checkManifest.dependsOn(processPackageMetadata)
+            try {
+                checkManifestProvider.configure { dependsOn(processPackageMetadata) }
+            } catch (e: NoSuchMethodError) {
+                @Suppress("DEPRECATION")
+                checkManifest.dependsOn(processPackageMetadata)
```
ChristianBecker: I fear we don't have another option besides relying on the `NoSuchMethodError` :(

plugin/build.gradle.kts
```diff
@@ -9,11 +9,7 @@ plugins {
 }
 
 dependencies {
-    // Using compileOnly because every android project applies
-    // its own version of the Android Plugin.
-    // We're staying on 3.0.1 so we don't accidentally use APIs
-    // that are not available in the client project
-    compileOnly("com.android.tools.build:gradle:3.0.1")
+    compileOnly("com.android.tools.build:gradle:3.3.0-alpha11")
```
ChristianBecker: `3.3.0-alpha12` was released yesterday. Or merge as-is and bump before release?

plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
SUPERCILEX: Yeah, I couldn't think of something else either.

plugin/build.gradle.kts
SUPERCILEX: I'll bump now

PR #424, opened 2018-10-11
Fix crash moving draft to completed stage

SUPERCILEX:


plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PromoteRelease.kt
```diff
@@ -43,6 +43,15 @@ open class PromoteRelease : PlayPublishPackageBase() {
                     updateFraction = extension._userFraction != null
             )
         }
+
+        // Duplicate statuses are not allowed, so only keep the unique ones from the highest version
+        // code.
+        track.releases = track.releases.sortedByDescending {
```
ChristianBecker: Do we actually need the `sortedByDescending` call?

SUPERCILEX: Probably not, but better be safe than sorry, right? 😂 We don't want to accidentally be removing higher version code releases.

ChristianBecker: You're totally right. Even if the `status` is the same, the other properties can be different. And then we want to keep the highest version code releases. 🤦‍♂️

PR #425, opened 2018-10-11
Short circuit image uploads if remote and local contents of graphics for type are identical

### LINKED ISSUE #423: Short circuit image uploads when hashes are the same
We currently upload images anytime the task runs, but this comes with issues if different machines who aren't sharing a Gradle cache try publishing. The same images will be uploading twice! This is bad because the Play Store cache is invalidated and forces app users to re-download those assets.

I don't think short circuiting text based uploads is worth it.

SUPERCILEX:
Fixes #423


SUPERCILEX:
@gtcompscientist No, it only downloads metadata (~250 bytes).

PR #426, opened 2018-10-11
Finally bootstrap high res graphics

SUPERCILEX:
I found a hack! Yes!!!


SUPERCILEX:
@gtcompscientist I wrote about it in my article (completely forgot to post in our Slack): https://proandroiddev.com/gradle-play-publisher-43abf77db344. Anyway, I randomly tried it on the URLs we get back from the API and it worked! Guess and check for the win. 😉😭

ChristianBecker:
I'm unsure about using such a hack (manipulating the returned URL by appending a constant string. This will break for sure if Google changes even a little bit. Can we eventually integrate a fallback to use the original URL if the manipulated URL results in a 404 NOT FOUND (or similar)?

On the other hand this will only affect the bootstrapping. The actual publishing process shouldn't fail if the bootstrapping fails, am I right?

SUPERCILEX:
@ChristianBecker This hack has been around [since 2016](https://stackoverflow.com/a/38907132/4548500), but your point seems valid so I fixed it.

PR #428, opened 2018-10-11
Circumvent task name conflicts by prefixing names with "gpp" should one occur

### LINKED ISSUE #427: maven-publish plugin task clash
### Describe the bug
I am also using the maven-publish plugin and get the following error;

A problem occurred evaluating script.
> Failed to apply plugin [class 'org.gradle.api.publish.plugins.PublishingPlugin']
   > Cannot add task ':app:publish' as a task with that name already exists.
apply plugin: 'maven-publish'

### To Reproduce

Add `apply plugin: 'maven-publish'`

#### Versions

 - Gradle Play Publisher: 2.0.0.beta 1
 - Gradle Wrapper (if applicable):
 - Android Gradle Plugin (if applicable):

### Expected behavior

I was able to use both the gradle-play-publisher 1.x version with the maven plugin in my gradle builds. When i upgrade to 2.0.0 beta 1 there is a clash.

SUPERCILEX:
Fixes #427

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Plugins.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Plugins.kt
@@ -36,10 +36,11 @@ internal inline fun <reified T : Task> Project.newTask(
         block()
     }
 
+    val safeName = if (tasks.findByName(name) == null) name else "gpp" + name.capitalize()
     return if (isConfigurationAvoidanceSupported) {
-        tasks.register(name, config)
+        tasks.register(safeName, config)
     } else {
-        val task = task(name, config)
+        val task = task(safeName, config)
         DefaultProvider { task }
     }
 }


plugin/src/main/kotlin/com/github/triplet/gradle/play/internal/Plugins.kt
```diff
@@ -36,10 +36,11 @@ internal inline fun <reified T : Task> Project.newTask(
         block()
     }
 
+    val safeName = if (tasks.findByName(name) == null) name else "gpp" + name.capitalize()
```
SUPERCILEX: TODO(me): we shouldn't be using `findByName` because it forces the task to be created.

SUPERCILEX: On second thought, we're _expecting_ these tasks to be null. Using `named` means we have to throw and catch an exception every time we want to create a new task.

PR #434, opened 2018-10-15
Remove unnecessary exclusion

SUPERCILEX:


SUPERCILEX:
This change will have to wait until our minimum supported AGP version is 3.1+.

PR #439, opened 2018-10-18
Remove deprecated tasks

SUPERCILEX:
@bhurling I don't remember, when had we decided to remove the deprecated tasks? Were we really planning on waiting for 3.0? Since we already shipped a release with the deprecated tasks, I'd vote to remove them for the final 2.0 release. WDYT?


bhurling:
Let's keep this one for 3.0. I wouldn't consider the previous beta a stable release. Especially since we found some bugs in there

SUPERCILEX:
Sounds fair 👍

Issue #477, opened 2018-12-14
Include Timeouts values in the Extension


cdsap:
Where our company is located, sometimes we experience problems of connectivity and the Timeouts are very short. 

I would like to have the possibility to update the values of the Timeouts on the Extension

SUPERCILEX:
See my comment on the PR, too much configuration is bad.

PR #478, opened 2018-12-14
Adding read and connect timeouts in the extension

cdsap:
Simple PR allowing update the values of the read/connect timeout through the Extension
#477


SUPERCILEX:
Thanks for submitting a PR, but please check with us first. I don't want to clutter the extension with pointless options (it's already getting out of hand), so I'd much rather simply increase the timeout.

I think 5 mins is fair. Please simply bump the timeout to `300_000`.

PR #479, opened 2018-12-14
Increase network timeout to 5 mins

cdsap:
following comment https://github.com/Triple-T/gradle-play-publisher/pull/478#issuecomment-447221229


SUPERCILEX:
Our dev with merge access is probably going to be out until January, so there might be a delay. Sorry in advance!

PR #492, opened 2019-01-13
Parallelize bootstrap image downloads

SUPERCILEX:
The worker API is actually available as far back as 3.1.

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayPublishTaskBase.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayPublishTaskBase.kt
@@ -23,7 +23,7 @@ abstract class PlayPublishTaskBase : DefaultTask(), ExtensionOptions {
     @get:Nested override lateinit var extension: PlayPublisherExtension
     @get:Internal internal lateinit var variant: ApplicationVariant
 
-    protected val workerExecutor = project.serviceOf<WorkerExecutor>()
+    @get:Internal protected val workerExecutor = project.serviceOf<WorkerExecutor>()
     @get:Internal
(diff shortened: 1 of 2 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayPublishTaskBase.kt
```diff
@@ -10,11 +10,14 @@ import org.gradle.api.tasks.Internal
...
 import org.gradle.internal.logging.progress.ProgressLogger
 import org.gradle.internal.logging.progress.ProgressLoggerFactory
+import org.gradle.kotlin.dsl.support.serviceOf
+import org.gradle.workers.WorkerExecutor
 
 abstract class PlayPublishTaskBase : DefaultTask(), ExtensionOptions {
     @get:Nested override lateinit var extension: PlayPublisherExtension
     @get:Internal internal lateinit var variant: ApplicationVariant
 
+    @get:Internal protected val workerExecutor = project.serviceOf<WorkerExecutor>()
```
bhurling: Do we need this property in our base task? The bootstrap task seems to be the only one using it

SUPERCILEX: At the moment, yes. I was thinking we might want to use it for future tasks though. I don't mind either way so if you want me to move it down, I'll be happy to do so.

bhurling: Let's move it into the Bootstrap task for now

SUPERCILEX: Done

Issue #517, opened 2019-02-16
Option to configure app bundle/apk name to publish


nlgtuankiet:
**Is your feature request related to a problem? Please describe.**
Recently I update my android gradle plugin from 3.5.0-alpha02 to 3.5.0-alpha04, they change the default app bundle name from `app.aab` to `app-release.aab`. This cause an error when run `gradlew publishReleaseBundle`:
```
File '...\app\build\outputs\bundle\release\app.aab' specified for property 'bundle' does not exist.
```

**Describe the solution you'd like**
Options to configure bundle file name or file path like:
```
play {
    bundlePath = "./build/outputs/bundle/release/app-release.aab"
}
```

SUPERCILEX:
Duplicate of #516

Issue #516, opened 2019-02-16
Breaking bundle file naming in AGP 3.5.0 alpha 04


knordbo:
### Describe the bug

Android Gradle Plugin 3.5.0-alpha04 changed bundle file names to include variant, similar to apks.

So what would previously be 'app.aab' is now 'app-release.aab' for 'release' variant

I suspected this was a bug all when AGP introduced bundles since the naming convention wasn't consistent with apks, sadly since they fixed it so late, it now breaks this plugin.

### To Reproduce

Run publish task (using bundles):

* What went wrong:
A problem was found with the configuration of task ':app:publishReleaseBundle'.
> File '../app/build/outputs/bundle/release/app.aab' specified for property 'bundle' does not exist.

#### Versions

 - Gradle Play Publisher: 2.1.0
 - Gradle Wrapper (if applicable): 5.2.1
 - Android Gradle Plugin (if applicable): 3.5.0-alpha04

gpx1000:
Annoying bug, tried to help, will downgrade gradle version used until it's fixed elsewhere.

SUPERCILEX:
Yeah, sorry again for shutting the PR down so fast. 😟 Adding new configurations options is almost never the answer. 🙂

knordbo:
Yeah I think the fact that google changed the path should not be exposed. Look for the file in the new path and fall back to the old way seems fine to me, and that piece of code can be deleted in the future.

gpx1000:
No worries, I agree that's probably not the best approach.  The first approach was to simply fix the bug by the if check (first commit).  Just simply do the if check.

gpx1000:
if the idea is to simply not have the option (completely agreed), should have just taken the first commit via cherry-pick or reverted the second commit.  It at least solves the immediate issue without waiting on a more elegant solution.  It's no worse then the already hard coded solution in code.

SUPERCILEX:
This will be fixed in v2.2.0.

SUPERCILEX:
Since v2.2.0 might be a ways away, I've decided to release a hotfix [v2.1.1](https://plugins.gradle.org/plugin/com.github.triplet.play/2.1.1) release.

PR #519, opened 2019-02-18
Allow for suffix addition on app bundles (i.e. app-release)

gpx1000:
issue #516 and #517


SUPERCILEX:
My apologies for not looking into this sooner, but this is not how we will solve the problem. I need to contact Xav to see if there is a way to not hardcode the value, otherwise we will check to see if the file exists or not.

Thanks for trying to help out!

PR #557, opened 2019-04-30
Fix bug causing app to be built even with artifact dir specified

### LINKED ISSUE #551: --artifact-dir doesn't work as expected
I'm using the latest build, which I believe is 2.2.0 at the current time.

I applied `--artifact-dir` to my build, but it didn't work, and it tried to start building my project. I didn't realize it at the time, but the directory I was pointing too was a relative path from the root project, not from the android module that I had the `play-publisher` plugin applied. It probably wasn't able to find it the directory, or the bundle I was trying to have it find, and instead of erring like I'd expect, instead it tried to build it.

I wasn't able to see an error, and actually understand what was happening until I tried `artifactDir`.

This behavior is unexpected. If I supply the `--artifact-dir` param, I'd expect it to behave the exact same as the `artifactDir` parameter.

SUPERCILEX:
Fixes #551

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
@@ -183,6 +183,23 @@ class PlayPublisherPlugin : Plugin<Project> {
                 generateBuildConfig.dependsOnCompat(processPackageMetadata)
             }
 
+            val publishApkTaskDependenciesHack = project.newTask(
+                    "publish${variantName}ApkWrapper",
+                    "",
+                    null
+            ) {
+                if (extension._artifactDir == null) {
+                    dependsOnCompat(processPackageMetadata)
+                    try {
+                        assembleProvider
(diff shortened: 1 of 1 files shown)


plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -183,6 +183,23 @@ class PlayPublisherPlugin : Plugin<Project> {
                 generateBuildConfig.dependsOnCompat(processPackageMetadata)
             }
 
+            val publishApkTaskDependenciesHack = project.newTask(
```
SUPERCILEX: Here's the bug: when using the global publish tasks, they will be resolved before their dependencies (aka the specific publish task) which means the artifact dir will have been resolved. However, this doesn't occur when using ./gradlew with the specific task directly because the artifact dir will be resolved after the task has already been initialized.

To get around this, we add a wrapper task that's in charge of lazily adding dependencies. This works because the real task gets its artifact dir resolved and then the task dependencies get initialized with access to the resolved artifact dir.

PR #579, opened 2019-05-22
Rewrite Play resource generator

### LINKED ISSUE #574: Rewrite listing language merging
The original reason for merging resources across languages was to support partial translation uploads for listings _only_. That is, you cannot upload a listing with the short/full description translated alone. The Play API will blow up and force you to specify a translated title even if it's the same as the native tongue one. This is true of each listing detail, but not of any other play metadata.

This reasoning has been lost over time which has led to the following bugs needing to be fixed:
- Graphics should not be merged at all. The Play Store handles default graphics on its own.
- If `fr-FR/full-description.txt` is added and the default language is `en-US`, publishing will fail because defaults aren't copied over.
- Updating a default language file doesn't update the untranslated one in incremental builds. That is, once the default file has been copied over, we think it's been translated manually and don't touch it anymore.

SUPERCILEX:
Fixes #574

### DIFF
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/GenerateResources.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/GenerateResources.kt
@@ -228,11 +228,11 @@ internal abstract class GenerateResources : DefaultTask() {
 
             parseSrcTree(prevIndex, prevReverseIndex, locales)
             buildIndex(index, reverseIndex, prunedResources)
-            pruneOutdatedReferences(index, reverseIndex, prunedResources)
-            mergeExistingReferences(prevIndex, prevReverseIndex, index, reverseIndex)
             insertNewLocales(index, reverseIndex, locales)
+            mergeExistingReferences(prevIndex, index, reverseIndex)
+            pruneOutdatedReferences(prevReverseIndex, index, reverseIndex, prunedResources)
             writeIndex(index, reverseIndex)
-            pruneGeneratedResources(prevIndex, prevReverseIndex, prunedResources)
+            pruneGeneratedResources(prevIndex, prevReverseIndex, index, prunedResources)
             generateResources(index)
         }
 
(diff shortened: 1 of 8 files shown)


play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/GenerateResources.kt
```diff
@@ -50,85 +52,47 @@ internal abstract class GenerateResources : DefaultTask() {
...
+            val areRootsValid = name == PLAY_PATH ||
+                    isDirectChildOf(PLAY_PATH) ||
+                    isChildOf(LISTINGS_PATH) ||
+                    isChildOf(RELEASE_NOTES_PATH) ||
+                    isChildOf(RELEASE_NAMES_PATH) ||
+                    isChildOf(PRODUCTS_PATH)
             check(areRootsValid) { "Unknown Play resource file: $this" }
+            check(extension != INDEX_MARKER) { "Resources cannot use the 'index' extension: $this" }
+            check(name != PLAY_PATH || !isChildOf(PLAY_PATH)) {
+                "The file name 'play' is illegal: $this"
```
tasomaniac: This was not a requirement in 2.6.x. 

In my app, I have a `flavor` called `play` and it was workin all fine until I upgraded to `2.7`. Should this change be mentioned in release notes?

And also, what is the reason really to keep it strict?

tasomaniac: Since I have `play` as a flavor. The `bootstrap` task also downloads stuff under `play/play` which then fails when I call `publish` task

SUPERCILEX: It would break write precedence: https://github.com/Triple-T/gradle-play-publisher/blob/edab9b6dacb1abd5db4472c018bc11b5ebc7700f/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/GenerateResources.kt#L441

But that's a bug, fixed in 4331aa614f676b2a839d50a403ef32e2c103685c.

PR #622, opened 2019-07-11
Changed task `publish` to `publishAll` to resolve conflict with core `maven-publish` plugin

### LINKED ISSUE #621: Conflict with maven-publish plugin
### Describe the bug

GPP conflicts with the core Gradle `maven-publish` plugin. The cause is that GPP uses the task name `publish` for publishing all artifacts.

### How To Reproduce

```
plugins {
    id 'com.github.triplet.play' version '2.2.1'
    id 'maven-publish'
}
```

#### Versions

- Gradle Play Publisher: 2.2.1
- Gradle Wrapper: 5.4.1
- Android Gradle Plugin: 3.5.0-beta05

#### Tasks executed

`publish`

### Expected behavior

To not throw an error message for an existing task.

### Additional context (if a crash, provide stack trace)

```
Gradle error:
Cannot add task 'publish' as a task with that name already exists.
```

ToxicBakery:
resolves #621 by changing the publish task name to publishAll matching its intent and use case while also resolving the conflict with the `maven-publish` plugin.

Worth noting I ran `./gradlew check` post changes

### DIFF
--- a/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
+++ b/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
@@ -58,7 +58,7 @@ class PlayPublisherPlugin : Plugin<Project> {
                 arrayOf(bootstrapOptionsHolder)
         )
         val publishAllTask = project.newTask<GlobalPublishableArtifactLifecycleTask>(
-                "publish",
+                "publishAll",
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
See https://github.com/Triple-T/gradle-play-publisher/issues/621#issuecomment-510541800. It'll get renamed to `gppPublish`.

SUPERCILEX:
Oh, right. The other reason this can't be done is because it breaks people using the task by name. The CLI is fine, but scripts will be broken.

PR #663, opened 2019-08-15
fixed enabled property name in readme example

ydanneg:
Based on this issue:
 Could not set unknown property 'isEnabled' for object of type com.github.triplet.gradle.play.PlayPublisherExtension.


SUPERCILEX:
This is normal. As you might have noticed, we have samples in both Kotlin and Groovy. The Kotlin property is `isEnabled` while the groovy one is `enabled`.

PR #689, opened 2019-09-07
Remove early property extraction

SUPERCILEX:
Requires Gradle 6.0


SUPERCILEX:
Will fix as part of #728

PR #736, opened 2019-11-14
Update deprecated GoogleCredential usage

pavlospt:
Hey there! 

Since we are using this great plugin in our app, I wanted to contribute something back :) 

This PR updates the deprecated usage of `GoogleCredential` class.

I have tested it with the command you added on `CONTRIBUTING.md` (`./gradlew tPublishBundle`).

Looking forward to your feedback!

### DIFF
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PlayPublishingApi.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PlayPublishingApi.kt
@@ -2,40 +2,49 @@ package com.github.triplet.gradle.play.tasks.internal.workers
 
 import com.github.triplet.gradle.common.utils.PLUGIN_NAME
 import com.github.triplet.gradle.play.PlayPublisherExtension
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.http.HttpRequestInitializer
 import com.google.api.client.http.javanet.NetHttpTransport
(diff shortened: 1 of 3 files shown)


play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PlayPublishingApi.kt
```diff
@@ -2,40 +2,49 @@ package com.github.triplet.gradle.play.tasks.internal.workers
...
                 .build()
     }
 
-    return AndroidPublisher.Builder(transport, JacksonFactory.getDefaultInstance()) {
-        credential.initialize(it.setReadTimeout(0))
-    }.setApplicationName(PLUGIN_NAME).build()
+    return AndroidPublisher.Builder(
+            buildNetTransport(),
+            JacksonFactory.getDefaultInstance(),
+            HttpRequestInitializer { credential.refresh() }
```
pavlospt: Not 100% sure if this is correct, but the Google Auth Library mentioned to refresh the Access token before using it!

pavlospt:
I believe the CI failure is irrelevant to the PR!

SUPERCILEX:
Thanks for taking initiative, but unfortunately, this change isn't desired:
- The CI failure is real and caused by an incompatible Guava dependency. Typically things will still work if you just exclude the `com.google.guava` module from the google api libs.
- The new Google auth doesn't support PKCS12 keys which means we can't switch over to the new stuff until a breaking v3 version of GPP is released. I want to wait a bit longer before doing that.
  - PS: your code compiles because `serviceAccountCredentials` is a property of `PlayPublisherExtension.Config`, but it won't work at runtime.

pavlospt:
Cool! Thanks for letting me know!

PR #799, opened 2020-04-01
Fixed typo

nicolas-raoul:


SUPERCILEX:
It was actually intentional, but I guess it doesn't have the connotation I thought it did: https://www.urbandictionary.com/define.php?term=copypasta&amp=true

nicolas-raoul:
Copypasta sounds tasty indeed :-)
Haha I double-checked the same urbandictionary page before submitting this PR.
Thanks for GPP, it saves so much time! 92 releases with it since 2015.

SUPERCILEX:
🙌

Issue #842, opened 2020-08-13
Wrong service account used for flavor


WhatDo:
### Describe the bug

I have multiple white label apps that go to different Play accounts, and i have a service account for each account.
I set the service account file for each flavor, and debugging the plugin i can see that the merged `PlayPublisherExtension` is correct, however, when the `GenerateEdit` task is run, the extension object has gotten its `serviceAccountCredentials` overwritten.

Output of the tasks is a 403 from the Play API since the wrong service account is being used.
Furthermore, i successfully called the Play API using the google python lib to confirm that the service account is correctly set up. 

### How To Reproduce

My build.gradle.kts looks roughly like

```kotlin
android {
  productFlavors {
    create("myflav1") { /* .. */ }
    create("myflav2") { /* .. */ }
  }

  playConfigs {
    val sa1 = file("mysa1.json")
    val sa2 = file("mysa2.json")
    register("myflav1") {
      serviceAccountCredentials.set(sa1)
    }

    register("myflav2") {
      serviceAccountCredentials.set(sa2)
    }
  }
}
```

#### Versions

- Gradle Play Publisher: 3.0.0-SNAPSHOT
- Gradle Wrapper: 6.6
- Android Gradle Plugin: 4.1.0-rc01

#### Tasks executed

`bootstrap` and `bootstrapMyflav1`

### Expected behavior

For my resources to be bootstrapped

### Additional context (if a crash, provide stack trace)

Add any other context about the problem here. If this bug is a crash, run the task with
`--stacktrace` to get the full context.

SUPERCILEX:
Was able to repo the bug. It's nasty, thanks for finding it!

SUPERCILEX:
Would you mind checking that it's been fixed in the 3.0 snapshots? (You might have to `./gradlew --refresh-dependencies`.)

WhatDo:
seems to work perfectly, thanks a lot

PR #843, opened 2020-08-13
Create new extension and fill it with config instead of overriding

WhatDo:
A shitty hack i made to make #842 work.

Problem seems to be that `mergeExtensions` always modifies the first entry in the list and returns it, but that means the last call in `android.onVariants` will become the common extension for all variants when any task is actually run.


SUPERCILEX:
The tests are failing so I can't accept this PR. I'll take a look at the bug at some point though.

WhatDo:
Yea it was mostly to illustrate what i found the problem to be. I've never developed a gradle plugin before, so i don't really know how to fix it the best way

SUPERCILEX:
Closing in favor of #844

PR #971, opened 2021-06-26
build: increase gradle memory, bump gradle to 7.1

mikehardy:
Hi there!

This is apologetically lame, but it's actually a PR made to demonstrate a problem, not to actually fix anything, but the goal is to be helpful

After all, what better reproduction than a project's own test suite :-)

This should obviously not be merged until a root cause is identified, but the hope is that since you are certainly more familiar with gradle and task dependencies than I am, you likely will see this and go "oh, just need to make this task depend on that task" or similar, where my solution process would start with "okay, so go read the gradle manual..." :sweat_smile: 

Here is an example of task failure output from a local `./gradlew test` run of this repo, and it is the exact same that [we see in AnkiDroid when we attempt to use GPP here along with gradle 7.1](https://github.com/ankidroid/Anki-Android/runs/2911340288?check_suite_focus=true#step:10:108):

```
org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in /tmp/junit15336675353018982252/app with arguments [--build-cache, installReleasePrivateArtifact, -S]

Output:
> Task :uploadReleasePrivateApk FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':uploadReleasePrivateApk'.
> Querying the mapped value of property(org.gradle.api.file.Directory, property(org.gradle.api.file.Directory, property(org.gradle.api.file.Directory, map(org.gradle.api.file.Directory flatmap(provider(task 'packageRelease', class com.android.build.gradle.tasks.PackageApplication)) check-type())))) before task ':packageRelease' has completed is not supported

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':uploadReleasePrivateApk'.
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
[...]

PR #972, opened 2021-06-26
build: automatically keep gradle up to date

mikehardy:
On the AnkiDroid project, we have integrated [this action](https://github.com/gradle-update/update-gradle-wrapper-action) to automagically keep our gradle up to date, and it partially explains how we got ahead of GPP here with regard to what gradle versions work

I'm happy to try to help here by carrying issues to this repo, e.g. #971 but I think it would be better if the compatible-canary PRs were running here vs just in our repo, so this seemed useful.

We've been using it since September 2020 with success.

Here is an example of it's work in our repo: https://github.com/ankidroid/Anki-Android/pull/9109
Here is where we integrated it: https://github.com/ankidroid/Anki-Android/pull/7182

Our current implementation: https://github.com/ankidroid/Anki-Android/blob/master/.github/workflows/update-gradle-wrapper.yml

I'm not sure about the repo token part, the distributions SHA thing [was important for us](https://github.com/ankidroid/Anki-Android/pull/7288) so I'm proposing it here, but the token part is no longer documented, may not be necessary so I'm not including it

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,18 @@
+name: Update Gradle Wrapper
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
Will take a look later.

SUPERCILEX:
Whoops, ignore the bot.

PR #1011, opened 2021-09-28
Update README.md (adds example of how to use `update` flag)

waliahimanshu:
Adds docs example to increase the rollout of existing in-progress release using `update` flag!


SUPERCILEX:
Thanks!

Issue #885, opened 2020-11-13
Could not get unknown property 'ReleaseStatus' for extension 'play'


kioli:
### Describe the bug

I've recently migrated from 2.3.0 to 3.0.0 and I changed my build.gradle file accordingly going from this
```
play {
    track = "beta"
    releaseStatus = "inProgress"
    userFraction = 1
    serviceAccountCredentials = file("credentials.json")
}
```
to this
```
play {
    track.set("beta")
    userFraction.set(new Double(1))
    releaseStatus.set(ReleaseStatus.IN_PROGRESS)
    serviceAccountCredentials.set(file("credentials.json"))
}
```
But when I try to build my app I am presented with the following error

`Could not get unknown property 'ReleaseStatus' for extension 'play' of type com.github.triplet.gradle.play.PlayPublisherExtension.`

### How To Reproduce

Include the `play` closure to the `build.gradle` file and sync

#### Versions

- Gradle Play Publisher: 3.0.0
- Gradle Wrapper:  6.5-all
- Android Gradle Plugin: 4.1.1

#### Tasks executed

It fails when syncing after modifying the `play` closure

### Expected behavior

I would expect the sync to be successful

SUPERCILEX:
Hmmm, looks like I need to update the docs for this. Someone else just asked the same thing: https://github.com/Triple-T/gradle-play-publisher/issues/827#issuecomment-725673682

kioli:
Right I forgot, I also use the plugin block
```
plugins {
    id 'com.android.application'
    id 'kotlin-android'
    id 'kotlin-android-extensions'
    id 'kotlin-kapt'
    id 'com.google.firebase.appdistribution'
    id 'com.google.gms.google-services'
    id 'com.github.triplet.play'
}
```

SUPERCILEX:
Invalidate caches and restart? 😝 If it works in Gradle, that's an IntelliJ problem and not related to this project.

kioli:
It doesn't solve it unfortunately
But all I can say so far is that it just wrong, but it does work 👍 
Thanks for your help and your great work

SUPERCILEX:
Bummer, hopefully an update to Studio or IntelliJ will fix it.

SUPERCILEX:
Note to self: also add comment about user fraction needing a d in groovy.

underwindfall:
I know this issue has been closed. But about the `red lint` was raised here , properly could be resovled by something like this 
```groovy
    play {
[...]

PR #1038, opened 2022-02-16
Expose `android-publisher` project in the Gradle plugin

JavierSegoviaCordoba:
Based on the issue https://github.com/Triple-T/gradle-play-publisher/issues/885, applying the Gradle plugin is not enough to be able to fully configure it (at least in conventional plugins).

The reason is there are some necessary APIs in `android-publisher` that are not exposed directly by the Gradle plugin, forcing the consumer to not only apply the Gradle plugin, also the consumer needs to add the `android-publisher` artifact.

With this change that requirement is no longer necessary.

### DIFF
--- a/play/plugin/build.gradle.kts
+++ b/play/plugin/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 }
 
 dependencies {
-    implementation(project(":play:android-publisher"))
+    api(project(":play:android-publisher"))
     implementation(project(":common:utils"))
     implementation(project(":common:validation"))
 


SUPERCILEX:
That's intentional: the APIs (even those not marked `internal`) are not designed for public consumption because it would be impossible to please everyone, they just happen to be what GPP needs. You are free to use that artifact, but I make no backwards compatibility guarantees (hence why it's not included by default).

Instead, I recommend using the android publisher API directly if you have advanced logic: https://github.com/googleapis/google-api-java-client-services/tree/main/clients/google-api-services-androidpublisher/v3. And of course, you can copy this repo's code as a starting point.

JavierSegoviaCordoba:
@SUPERCILEX why `ReleaseStatus` is not designed for public consumption?

SUPERCILEX:
Oh, didn't realize that's what this PR was getting at. I created #1039 with some ideas to fix it properly.

Issue #1068, opened 2022-10-12
Update Android Gradle Plugin to 7.30


tanman2391:
### Problem description
upgrade request to be compatible with latest Android Gradle Plugin

andreimesina:
Hey, any updates on this one? Lots of projects will soon be on Android Gradle Plugin 7.3.+

SUPERCILEX:
I'm no longer actively maintaining this project, so someone else will need to take this on. There was an attempt, but it got stalled because gradle broke things as usual: https://github.com/Triple-T/gradle-play-publisher/pull/1069.

PR #1069, opened 2022-10-12
updating agp version to 7.3.0

tanman2391:
Hey there! So you want to contribute to Gradle Play Publisher?
Before you file this pull request, follow these steps:

- Read [the contribution guidelines](https://github.com/Triple-T/gradle-play-publisher/blob/master/CONTRIBUTING.md).
- If this has been discussed in an issue, make sure to mention the issue number here.
  If not, go file an issue about this to make sure it is a desirable change.

https://github.com/Triple-T/gradle-play-publisher/issues/1068

### DIFF
--- a/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
+++ b/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
@@ -231,8 +231,8 @@ internal class DefaultPlayPublisher(
                     println("Uploading $thing: ${(it.progress * 100).roundToInt()}% complete")
                 MediaHttpUploader.UploadState.MEDIA_COMPLETE ->
                     println("${thing.capitalize()} upload complete")
-                MediaHttpUploader.UploadState.NOT_STARTED -> TODO()
-                MediaHttpUploader.UploadState.INITIATION_COMPLETE -> TODO()
+                MediaHttpUploader.UploadState.NOT_STARTED -> {}
+                MediaHttpUploader.UploadState.INITIATION_COMPLETE -> {}
             }
         }
         return this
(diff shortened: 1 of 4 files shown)


SUPERCILEX:
Oh, I had a branch that fixed the compilation issue.

SUPERCILEX:
Hmmm, actually I can't find it anymore. You'll just need to add the other two cases and have them do nothing.

tanman2391:
> Hmmm, actually I can't find it anymore. You'll just need to add the other two cases and have them do nothing.

done. I added simple println calls for the two cases, as I dont like empty flows and didn't want to add a catch all else branch.

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
```diff
@@ -231,6 +231,8 @@ internal class DefaultPlayPublisher(
                     println("Uploading $thing: ${(it.progress * 100).roundToInt()}% complete")
                 MediaHttpUploader.UploadState.MEDIA_COMPLETE ->
                     println("${thing.capitalize()} upload complete")
+                MediaHttpUploader.UploadState.NOT_STARTED -> TODO()
+                MediaHttpUploader.UploadState.INITIATION_COMPLETE -> TODO()
```
SUPERCILEX, line 235: Nope, please do this: `-> {}`

SUPERCILEX: Oh gotcha. Yeah, I prefer not printing stuff since this is just supposed to show upload progress and we don't really care about the details (those are already logged if you enable Gradle debug logging).

SUPERCILEX:
Looks like you'll need to check the Gradle release notes and see what they changed about the pom. Gradle upgrades are unpleasant, sorry.

saiday:
@tanman2391 @SUPERCILEX Thank you for your effort.

Issue #1083, opened 2023-04-16
Configuration Cache issue with Gradle 8.1


eygraber:
After updating to Gradle 8.1 I'm getting a configuration cache error:

```
Task `:commitEditFor<my package>` of type `com.github.triplet.gradle.play.tasks.CommitEdit`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/8.1/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

There was no error when using Gradle 8.0

eygraber:
Bump

SUPERCILEX:
PRs welcome, but I don't work on this project anymore.

ydanneg:
any plans to release it?

gmazzo:
Reponing since #1084 was not a fix at all.

gmazzo:
This issue should be reopned since #1084 was not a fix at all. I'm taking a look at it

SUPERCILEX:
@gmazzo Can you confirm the fix with the snapshot builds? If not, no worries I'll just publish.

gmazzo:
> @gmazzo Can you confirm the fix with the snapshot builds? If not, no worries I'll just publish.

Give me a couple of days and I'll report back
cc: @rolgalan

gmazzo:
> > @gmazzo Can you confirm the fix with the snapshot builds? If not, no worries I'll just publish.
> 
> Give me a couple of days and I'll report back cc: @rolgalan

I can confirm it worked successfully with CC on for `publishReleaseBundle` and `promoteArtifact` 🚀 

![image](https://user-images.githubusercontent.com/513566/236841344-c5d34392-b0ff-418c-9fb0-f70bf227f31c.png)
![image](https://user-images.githubusercontent.com/513566/236841835-9537b8a5-3e9f-4ec9-88e3-88b32799c946.png)

gmazzo:
> > @gmazzo Can you confirm the fix with the snapshot builds? If not, no worries I'll just publish.
> 
> Give me a couple of days and I'll report back cc: @rolgalan

I can confirm it worked successfully with CC on for `publishReleaseBundle` and `promoteArtifact` 🚀 

![image](https://user-images.githubusercontent.com/513566/236841344-c5d34392-b0ff-418c-9fb0-f70bf227f31c.png)
[...]

PR #1084, opened 2023-04-24
Fixed issues with Configuration Cache

gmazzo:
Fixes https://github.com/Triple-T/gradle-play-publisher/issues/1083 by removing the `Task.project` reference inside the `onlyIf` closure of `CommitEdit`

### DIFF
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
@@ -4,6 +4,7 @@ import com.github.triplet.gradle.play.PlayPublisherExtension
 import com.github.triplet.gradle.play.tasks.internal.PublishTaskBase
 import com.github.triplet.gradle.play.tasks.internal.workers.PlayWorkerBase
 import com.github.triplet.gradle.play.tasks.internal.workers.paramsForBase
+import org.gradle.api.invocation.Gradle
 import org.gradle.api.logging.Logging
 import org.gradle.api.tasks.TaskAction
 import org.gradle.kotlin.dsl.submit
@@ -11,14 +12,17 @@ import org.gradle.work.DisableCachingByDefault
 import org.gradle.workers.WorkerExecutor
 import javax.inject.Inject
 
+@Suppress("LeakingThis")
 @DisableCachingByDefault
 internal abstract class CommitEdit @Inject constructor(
+        private val gradle: Gradle,
         extension: PlayPublisherExtension,
         private val executor: WorkerExecutor,
 ) : PublishTaskBase(extension) {
+
     init {
         onlyIf {
-            val buildFailed = project.gradle.taskGraph.allTasks.any { it.state.failure != null }
+            val buildFailed = gradle.taskGraph.allTasks.any { it.state.failure != null }
             if (buildFailed) apiService.get().cleanup()
             !buildFailed
         }


ZacSweers:
Are you sure `gradle` is safe to access here? Seems like this should be wrapped in a provider instead. A test would help

gmazzo:
> Are you sure `gradle` is safe to access here? Seems like this should be wrapped in a provider instead. A test would help

Provider of what exactly? `Provider<Gradle>`?

I didn't add all my changes (like bumping Gradle to 8.1.1 and AGP to 8.0.0, enabling Configuration Cache (CC) or setting `java.toolchain` DSL to `8`)  to reduce the PR to the bare minimum, but the task `CommitEdit` is getting instantiated as part the existing integrations tests, so it does work. Which kind of extra test do you have in mind?

Regarding keeping a reference to the `Gradle` object, as far I know, is not a violation of CC requirements. At least is not being notified as that by Gradle as it does with `Task.project`.

ZacSweers:
A provider of a Boolean (so compute the whole thing). A test that verifies this task does not break configuration cache (I.e. run it twice with the same inputs, confirm the second time reuses the cache).

gmazzo:
> A provider of a Boolean (so compute the whole thing). A test that verifies this task does not break configuration cache (I.e. run it twice with the same inputs, confirm the second time reuses the cache).

A few things:
1) Let's say I add a `onlyIf(buildHasNotFailedProvider::get)`, where should I move the logic for computing its valid, to the plugin? I personally try to avoid self config blocks with `init` at all, as that logic should belong to the plugin. In this case, not sure if the refactor will be justified as (and if I understood the logic correct) this seems to be to abort the transaction on Google Play in case any other previous task has failed. If that's the case, probably this logic is a bit hacky and it should be replaced with a `finalizedBy` instead. I feel changing this will be a bit unrelated to the purpose of this PR, which is just make it compatible with CC.

2) CC is not exactly the same than Build Cache, and not so easily to test. CC will fail depending on the tasks that were effectively run. So to make sure the whole plugin is compatible with it, I'd just enable CC on the tests as well. 
I tried that, and it was working fine, but later the `testapp` uses another `version-orchestrator` plugin which is also incompatible with CC, making the test to fail. 
That was when I decided to limit the PR to the bare minimum to fix this issue.

3) Regarding this:
> I.e. run it twice with the same inputs, confirm the second time reuses the cache

As far I know, and I may incorrect, there is no API (yet?) from `GradleRunner` tests to tell if CC was a hit or not. There is one for BuildCache from TaskOutput API, but no for this one.  What kind of test would you have in mind?

I decide to work on this because it's actually blocking us from updating to Gradle 8.1.1 as it will break on release automation. I'm afraid going deeper on the refactor, bumping/changing plugins because they are also incompatible may extend too much in time and potentially introduce new issues.

SUPERCILEX:
This thing is a hack because CommitEdit *is* a finalizer and gradle has no API to detect whether or not the finalized tasks have failed. The goal is to prevent publishing if something went wrong.

It's a bummer if this break CC, but we should probably make the plugin work on Gradle 8.1, so gonna merge.

gmazzo:
Thankyou!

JavierSegoviaCordoba:
@SUPERCILEX there is a new scope APIs, if I remember correctly, `FlowScope`.

SUPERCILEX:
> @SUPERCILEX there is a new scope APIs, if I remember correctly, `FlowScope`.

If someone wants to give that a whirl, I'd be happy to accept PRs.

jrgonzalezg:
This does not seem to work for me. The tasks still fails but now it has a different error message:
```
Task .... of type `com.github.triplet.gradle.play.tasks.CommitEdit`: cannot serialize object of type 'org.gradle.invocation.DefaultGradle', a subtype of 'org.gradle.api.invocation.Gradle', as these are not supported with the configuration cache. See https://docs.gradle.org/8.1.1/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```

gmazzo:
Damn, I thought I tested that. If you are seeing that error message then
it's not going to work.

gmazzo:
I can take a look at this early next week, this time with a test

ZacSweers:
> Are you sure `gradle` is safe to access here? Seems like this should be wrapped in a provider instead. A test would help

^^ I wasn't trying to be annoying, this is exactly why I suggested it.

PR #1086, opened 2023-05-02
Fixes `CommitEdit` issues when enabling ConfigurationCache

gmazzo:
Follow up PR of https://github.com/Triple-T/gradle-play-publisher/pull/1084 partially addressing #1083 

In order to work con `Gradle 8.1.1` Configuration Cache issues, we need to bump Gradle, AGP to `8.0.0` and set the minimum JDK to 11 (17 for running the tests).
All these changes were done in the first commit. 

Focussing exclusively on `CommitEdit` task, I've made some changes as @ZacSweers suggested but still I didn't manage to get it to work as `onlyIf` seems to be computed at configuration time ([asked in Slack if this is expected, probably will fill an issue](https://gradle-community.slack.com/archives/C013WEPGQF9/p1682304058131779))

Said that, there are still many tests broken if we enable CC for the whole test suite, so being fully compatible with CC is going to be a more complex task. If any is eager to contribute, I've added a `withConfigurationCache` property to `IntegrationTestBase`'s constructor so we can start fixing them incrementally:
```kotlin
class CommitEditIntegrationTest : IntegrationTestBase(withConfigurationCache = true), SharedIntegrationTest
```

Once the migration is done, we can just remove this flag and turn it of by adding the properties into `play/plugin/src/test/fixtures/app/gradle.properties`:
```properties
org.gradle.configuration-cache=true
```

Here is the list of failing tests after this PR (if CC is enabled for all of them):
image

### DIFF
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
@@ -41,28 +41,37 @@ import com.github.triplet.gradle.play.tasks.internal.GlobalPublishableArtifactLi
 import com.github.triplet.gradle.play.tasks.internal.GlobalUploadableArtifactLifecycleTask
 import com.github.triplet.gradle.play.tasks.internal.PlayApiService
 import com.github.triplet.gradle.play.tasks.internal.PublishArtifactTaskBase
+import com.github.triplet.gradle.play.tasks.internal.PublishTaskBase
 import com.github.triplet.gradle.play.tasks.internal.PublishableTrackLifecycleTask
 import com.github.triplet.gradle.play.tasks.internal.UpdatableTrackLifecycleTask
 import com.github.triplet.gradle.play.tasks.internal.WriteTrackLifecycleTask
+import org.gradle.api.DomainObjectSet
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.plugins.ExtensionAware
 import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildServiceRegistration
+import org.gradle.build.event.BuildEventsListenerRegistry
 import org.gradle.kotlin.dsl.container
 import org.gradle.kotlin.dsl.create
+import org.gradle.kotlin.dsl.domainObjectSet
 import org.gradle.kotlin.dsl.findPlugin
 import org.gradle.kotlin.dsl.getByType
 import org.gradle.kotlin.dsl.invoke
+import org.gradle.kotlin.dsl.mapProperty
 import org.gradle.kotlin.dsl.named
 import org.gradle.kotlin.dsl.newInstance
 import org.gradle.kotlin.dsl.registerIfAbsent
+import org.gradle.kotlin.dsl.setProperty
 import org.gradle.kotlin.dsl.the
 import org.gradle.kotlin.dsl.withType
+import javax.inject.Inject
 
 @Suppress("unused") // Used by Gradle
-internal class PlayPublisherPlugin : Plugin<Project> {
+internal abstract class PlayPublisherPlugin @Inject constructor(
+        private val buildEventsListenerRegistry: BuildEventsListenerRegistry,
+) : Plugin<Project> {
     override fun apply(project: Project) {
         project.validateRuntime()
 
@@ -236,6 +245,8 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                 parameters.appId.set(appId)
                 parameters.editIdFile.set(project.layout.buildDirectory.file("$OUTPUT_PATH/$appId.txt"))
             }
+            buildEventsListenerRegistry.onTaskCompletion(api)
+
             project.gradle.sharedServices.registrations.named<
                     BuildServiceRegistration<PlayApiService, PlayApiService.Params>
                     >("playApi-$appId") {
@@ -248,6 +259,11 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                 }
             }
 
+            fun PublishTaskBase.bindApi(api: Provider<PlayApiService>) {
+                usesService(api)
+                apiService.value(api).disallowChanges()
+            }
+
             // ------------------- START: ALL BUILD TYPE VARIANT SPECIFIC TASKS -------------------
 
             val publishInternalSharingApkTask = project.newTask<PublishInternalSharingApk>(
@@ -258,7 +274,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                     """.trimMargin(),
                     arrayOf(extension, executionDir),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 apks.from(findApkFiles())
                 outputDirectory.set(project.layout.buildDirectory.dir(
                         "outputs/internal-sharing/apk/${variant.name}"))
@@ -276,7 +292,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                     """.trimMargin(),
                     arrayOf(extension, executionDir),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 bundles.from(findBundleFiles())
                 outputDirectory.set(project.layout.buildDirectory.dir(
                         "outputs/internal-sharing/bundle/${variant.name}"))
@@ -317,7 +333,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                     """.trimMargin(),
                     arrayOf(extension, bootstrapOptionsHolder),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 srcDir.set(project.file("src/${variant.flavorNameOrDefault}/$PLAY_PATH"))
             }
             bootstrapAllTask { dependsOn(bootstrapTask) }
@@ -355,7 +371,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                     """.trimMargin(),
                     arrayOf(extension, executionDir),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 resDir.set(resourceDir)
 
                 finalizedBy(commitEditTask)
@@ -370,7 +386,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                     """.trimMargin(),
                     arrayOf(extension),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 productsDir.setFrom(resourceDir.map {
                     it.dir(PRODUCTS_PATH).asFileTree.matching { include("*.json") }
                 })
@@ -391,7 +407,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
                             extension.resolutionStrategy.get() == ResolutionStrategy.AUTO_OFFSET,
                     ),
             ) {
-                apiService.set(api)
+                bindApi(api)
                 versionCodes.set(staticVersionCodes)
                 playVersionCodes.set(project.layout.buildDirectory.file(
                         "$INTERMEDIATES_OUTPUT_PATH/${variant.name}/available-version-codes.txt"))
@@ -406,7 +422,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
 
 
             fun PublishArtifactTaskBase.configureInputs() {
-                apiService.set(api)
+                bindApi(api)
                 releaseNotesDir.set(resourceDir.map {
                     it.dir(RELEASE_NOTES_PATH).also { it.asFile.safeMkdirs() }
                 })
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
@@ -6,26 +6,18 @@ import com.github.triplet.gradle.play.tasks.internal.workers.PlayWorkerBase
 import com.github.triplet.gradle.play.tasks.internal.workers.paramsForBase
 import org.gradle.api.invocation.Gradle
 import org.gradle.api.logging.Logging
+import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.TaskAction
 import org.gradle.kotlin.dsl.submit
 import org.gradle.work.DisableCachingByDefault
 import org.gradle.workers.WorkerExecutor
 import javax.inject.Inject
 
-@Suppress("LeakingThis")
 @DisableCachingByDefault
 internal abstract class CommitEdit @Inject constructor(
-        private val gradle: Gradle,
         extension: PlayPublisherExtension,
         private val executor: WorkerExecutor,
 ) : PublishTaskBase(extension) {
-    init {
-        onlyIf {
-            val buildFailed = gradle.taskGraph.allTasks.any { it.state.failure != null }
-            if (buildFailed) apiService.get().cleanup()
-            !buildFailed
-        }
-    }
 
     @TaskAction
     fun commit() {
@@ -36,14 +28,14 @@ internal abstract class CommitEdit @Inject constructor(
 
     abstract class Committer : PlayWorkerBase<Committer.Params>() {
         override fun execute() {
-            if (apiService.shouldCommit()) {
+            if (apiService.shouldCommit) {
                 println("Committing changes")
                 try {
                     apiService.commit()
                 } finally {
                     apiService.cleanup()
                 }
-            } else if (apiService.shouldSkip()) {
+            } else if (apiService.shouldSkip) {
                 println("Changes pending commit")
                 try {
                     apiService.validate()
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
@@ -83,9 +83,7 @@ internal abstract class PlayApiService @Inject constructor(
     }
 
     override fun onFinish(event: FinishEvent) {
-        if (event is TaskFinishEvent) {
-            buildFailed = buildFailed || (event.result is TaskFailureResult)
-        }
+        buildFailed = buildFailed || event.result is TaskFailureResult
     }
 
     private fun getOrCreateEditId(): String {
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
@@ -44,9 +44,6 @@ internal abstract class PlayApiService @Inject constructor(
     var buildFailed = false
         private set
 
-    var cleanupAtClose = true
-        get() = field && shouldCommit
-
     fun scheduleCommit() {
         editIdFile.marked("commit").safeCreateNewFile()
     }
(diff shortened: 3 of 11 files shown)


build.gradle.kts
```diff
@@ -63,24 +63,24 @@ allprojects {
     group = "com.github.triplet.gradle"
 
     afterEvaluate {
-        convention.findByType<JavaPluginExtension>()?.apply {
-            sourceCompatibility = JavaVersion.VERSION_1_8
+        extensions.findByType<JavaPluginExtension>()?.apply {
+            toolchain.languageVersion.convention(JavaLanguageVersion.of(11))
```
gmazzo, line 67: Gradle 8.1.1 seems to need AGP 8 (I've seen some `NoSuchMethodError` in the logs related to CC) and AGP 8 requires a minimum of JDK11. Actually, the plugin was already using [JDK11 API (readNBytes) ](https://github.com/Triple-T/gradle-play-publisher/blob/master/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/Bootstrap.kt#L187) but declared that it can run on JDK8.

play/plugin/build.gradle.kts
```diff
@@ -51,6 +51,11 @@ afterEvaluate {
 tasks.withType<Test> {
     inputs.files(fileTree("src/test/fixtures"))
 
+    // AGP 8 requires JDK 17 and we want to to be compatible with previous JDKs
+    javaLauncher.set(javaToolchains.launcherFor {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    })
```
gmazzo, line 57: Even AGP 8 can be consumed by JDK11, it has a hard check on runtime to require JDK17. I've workaround this by changing the toolchain on the `test` task. Bumping the whole build directly to JDK17 may affect compatibility with the consumers of the plugins.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/PlayPublisherPlugin.kt
```diff
@@ -173,7 +173,7 @@ internal class PlayPublisherPlugin : Plugin<Project> {
         (android as ExtensionAware).extensions.add(PLAY_CONFIGS_PATH, extensionContainer)
 
         val androidExtension = project.extensions.getByType<ApplicationAndroidComponentsExtension>()
-        androidExtension.onVariants v@{ variant ->
+        androidExtension.onVariants(androidExtension.selector().all()) v@{ variant ->
```
gmazzo, line 176: This is an undocumented breaking change of AGP 8 that I faced before. Because Kotlin defaultImpls works, if the plugin is compiled using AGP `compileOnly` dependency, the byte generated lack of some impl method causing `NoSuchMethodError`, when the consumer AGP version is lower than 8.

I'm not sure if there is any flag in the compiler plugin to avoid this issue, any suggestion is welcome :)

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
```diff
@@ -15,13 +15,16 @@ import javax.inject.Inject
...
 @DisableCachingByDefault
 internal abstract class CommitEdit @Inject constructor(
-        private val gradle: Gradle,
         extension: PlayPublisherExtension,
         private val executor: WorkerExecutor,
 ) : PublishTaskBase(extension) {
     init {
+        val buildFailedProvider = project.provider {
+            project.gradle.taskGraph.allTasks.any { it.state.failure != null }
+        }
```
gmazzo, line 24: This is not working to me @ZacSweers , based on my tests, the `[onlyIf` closure is evaluated at configuration time](https://gradle-community.slack.com/archives/C013WEPGQF9/p1683047848634369?thread_ts=1682304058.131779&cid=C013WEPGQF9) when it shouldn't be and of course the task outcome is not yet populated.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/GenerateResources.kt
```diff
@@ -52,6 +53,7 @@ internal abstract class GenerateResources @Inject constructor(
     abstract val resSrcDirs: ListProperty<Directory>
 
     @get:SkipWhenEmpty
+    @get:IgnoreEmptyDirectories
```
gmazzo, line 56: Something internally has changed on Gradle 8.1.1 and `IgnoreEmptyDirectories` now is required to keep getting `NO-SOURCE` outcome on the tests

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishListings.kt
```diff
@@ -84,6 +85,7 @@ internal abstract class PublishListings @Inject constructor(
     @Suppress("MemberVisibilityCanBePrivate", "unused")
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:SkipWhenEmpty
+    @get:IgnoreEmptyDirectories
```
gmazzo, line 88: Same than https://github.com/Triple-T/gradle-play-publisher/pull/1086/files#r1182848732

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/helpers/IntegrationTestBase.kt
```diff
@@ -13,7 +13,9 @@ import java.util.concurrent.BlockingQueue
 import kotlin.math.max
 import kotlin.random.Random
 
-abstract class IntegrationTestBase : IntegrationTest {
+abstract class IntegrationTestBase(
+        override val withConfigurationCache: Boolean = false
```
gmazzo, line 17: I've added this to allow fixing CC compatibility incrementally.

play/plugin/build.gradle.kts
```diff
@@ -59,23 +64,16 @@ tasks.withType<Test> {
...
         implementationClass = "com.github.triplet.gradle.play.PlayPublisherPlugin"
-    }
-}
-
-pluginBundle {
-    website = "https://github.com/Triple-T/gradle-play-publisher"
-    vcsUrl = "https://github.com/Triple-T/gradle-play-publisher"
-    tags = listOf("android", "google-play", "publishing", "deployment", "apps", "mobile")
-
-    mavenCoordinates {
```
gmazzo, line 76: I been forced to bump `gradlePublish` version to `1.1.0` to be compatible with Gradle 8 to address [this](https://github.com/Triple-T/gradle-play-publisher/actions/runs/4864010574/jobs/8672470785#step:7:174) and as far I know, the `mavenCoordinates` closure is not available anymore

gmazzo: `website`, `vcsUrl` and `tags` are now part of the  `gradlePlugin` closure

SUPERCILEX: What happened to `groupId` and `artifactId`? I feel like publishing will probably fail without them.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
gmazzo: @SUPERCILEX  how do you feel about making `CommitEdit` task a "success only" one and moving the `apiService.cleanup` logic to the `PlayApiService` by adding `AutoCloseable`?

From the [docs](https://docs.gradle.org/current/userguide/build_services.html):
> A build service implementation can also optionally implement AutoCloseable, in which case Gradle will call the build service instance’s close() method when it discards the service instance. This happens some time between completion of the last task that uses the build service and the end of the build.

I could let commit run only if all tasks calling it with `finalizedBy` also does `dependsOn` in the opposite way, as suggested [here](https://gradle-community.slack.com/archives/C013WEPGQF9/p1683045329792689?thread_ts=1682304058.131779&cid=C013WEPGQF9).

Later `PlayApiService` will be responsible of cancelling any pending transaction on its `close` method.

play/plugin/build.gradle.kts
gmazzo: From the [docs](https://plugins.gradle.org/docs/publish-plugin#breaking-changes):
> #### Breaking changes:
> The mavenCoordinates block is no longer defined and can’t be used to override publication GAV coordinates. The coordinates are always based on the Maven publication’s GAV (which defaults to the Project’s group, name & version).

gmazzo: I also noted that `publishToMavenLocal` is broken and generates a POM with an invalid `plugin` artifactId in its dependency. I can make it work by removing any explicit `artifactId` set in the POM configuration and replace the `plugin` project's name with `play-publisher` in the settings

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
SUPERCILEX: I think I had considered using AutoCloseable before but to handle both success and failure. I think using it just for failure cleanup makes sense, though I'm not clear on how you prevent the CommitEdit task from trying to commit if a task failed (can you paste whatever is in the slack?).

SUPERCILEX:
Taking a step back, I think we need a PR focussed on just upgrading dependencies/compat with AGP 8 etc. And then we can have PRs for the test changes + CC fixes.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/CommitEdit.kt
gmazzo: Sure, it's basically linked them both ways:
```kotlin
publishApkTasl.configure { finalizedBy(commitTask) }
commitTask.configure { dependsOn(publishApkTasl) }
```

But now I'm seeing another issue, we have 4 potential (if I remember correctly) tasks `finalizedBy`. We want `commit` to run if ANY of those has succeed, not to force all of them to run 🤔

gmazzo: I managed to get it to work by introducing `AutoCloseable` to `cleanup` and `OperationCompletionListener` to keep track of failed tasks on its new `tasksHasFailed` property.
Also added the `usesService` API call allowing the `PlayApiService` to close right after the last task using it is done.

The behavior seems to be the same, and tests are passing now (expect for one which I commented).

Can you review this now? thanks!

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/GenerateEditIntegrationTest.kt
```diff
@@ -72,6 +72,8 @@ class GenerateEditIntegrationTest : IntegrationTestBase(), SharedIntegrationTest
                             .named("playApi-com.example.publisher")
                             .get().service.get() as PlayApiService
 
+                service.cleanupAtClose = false
```
gmazzo, line 75: Added `cleanupAtClose` just for this test, as it's checking the `editFile` after the build is gone. As now the service closes before the build ends, this file is getting deleted.

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/shared/PublishArtifactIntegrationTests.kt
```diff
@@ -74,7 +74,7 @@ interface PublishArtifactIntegrationTests : SharedIntegrationTest {
         assertArtifactUpload(result1)
 
         result2.requireTask(outcome = SUCCESS)
-        assertThat(result2.output).contains("didPreviousBuildSkipCommit=true")
+        assertThat(result2.output).contains("didPreviousBuildSkipCommit=false")
```
gmazzo, line 77: I'm not fully getting what this test suposted to do. It started to fail after introducing the changes on `PlayApiService` (probably the `AutoCloseable` + `usesService`).

Based on how it looks, should `play.commit` in line 66 be `true` already?  

Anyway, `didPreviousBuildSkipCommit` seems now to be always false, which as far I understand, makes sense, since there no possibility the that the previous build skiped the commit, since it's going to be cancelled anyway at the end when `AutoCloseeable.close` is called.

Should `commit = false` imply `cleanupAtClose = false` aswell?

SUPERCILEX:
> I think we need a PR focussed on just upgrading dependencies/compat with AGP 8 etc. \

Can we still do this? Or am I misunderstanding and upgrading requires fixing the CC issue?

gmazzo:
> > I think we need a PR focussed on just upgrading dependencies/compat with AGP 8 etc.
> 
> Can we still do this? Or am I misunderstanding and upgrading requires fixing the CC issue?

I have splited already in two commits targeting this, I'll open another PR with the first one, yes

gmazzo:
> > I think we need a PR focussed on just upgrading dependencies/compat with AGP 8 etc.
> 
> Can we still do this? Or am I misunderstanding and upgrading requires fixing the CC issue?

There:
https://github.com/Triple-T/gradle-play-publisher/pull/1087

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/shared/PublishArtifactIntegrationTests.kt
gmazzo: > Should commit = false imply cleanupAtClose = false as well?
I'm going to assume that for now, so the test will pass. I think it makes sense.

SUPERCILEX: Yes, commit false allows you to keep a pending publishing action across builds.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
```diff
@@ -69,6 +79,18 @@ internal abstract class PlayApiService @Inject constructor(
...
 
+    override fun close() {
+        if (cleanupAtClose) {
+            cleanup()
+        }
+    }
+
+    override fun onFinish(event: FinishEvent) {
+        if (event is TaskFinishEvent) {
+            tasksHasFailed = tasksHasFailed || (event.result is TaskFailureResult)
```
SUPERCILEX, line 90: |=

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
```diff
@@ -37,11 +41,17 @@ internal abstract class PlayApiService @Inject constructor(
     private val editIdFileAndFriends
         get() = listOf(editIdFile, editIdFile.marked("commit"), editIdFile.marked("skipped"))
 
+    var tasksHasFailed = false
```
SUPERCILEX, line 44: Can you call this `buildFailed`? If that conflicts, then maybe `buildHasFailingTasks`.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
gmazzo: Yeah I know, but it does not exist in Kotlin:
https://kotlinlang.org/docs/keyword-reference.html#operators-and-special-symbols

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/GenerateEditIntegrationTest.kt
gmazzo: Actually I just needed to add `usesService(service)` here and not `service.cleanupAtClose = false`

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
```diff
@@ -37,11 +41,17 @@ internal abstract class PlayApiService @Inject constructor(
     private val editIdFileAndFriends
         get() = listOf(editIdFile, editIdFile.marked("commit"), editIdFile.marked("skipped"))
 
+    var buildFailed = false
+        private set
+
+    var cleanupAtClose = true
```
SUPERCILEX, line 47: This isn't needed anymore right? Can just be replaced with `shouldCommit` down in `fun close`.

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/shared/PublishArtifactIntegrationTests.kt
```diff
@@ -63,7 +63,7 @@ interface PublishArtifactIntegrationTests : SharedIntegrationTest {
         // language=gradle
         val config2 = """
             android.defaultConfig.versionCode 2
-            play.commit = false
+            play.commit = true
```
SUPERCILEX, line 66: I'm looking at this again and I'm not sure it's right. Does it still fail if you go back to false?

SUPERCILEX (commented):
Almost there! Thanks for all the hard work.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
gmazzo: Very true, it's gone now 💪

play/plugin/src/test/kotlin/com/github/triplet/gradle/play/tasks/shared/PublishArtifactIntegrationTests.kt
gmazzo: The test now passes with both values, so I misinterpreted the test's meaning. Reverted

SUPERCILEX: Perfect!

SUPERCILEX (approved):

SUPERCILEX:
@ZacSweers Any opinions?

ZacSweers:
I must confess this feels like a roundabout solution and likely to have issues in the future. Tasks shouldn't run based on the status of the entire build, they should usually just run with either explicit task dependencies or finalizedBy on the relevant tasks. Either that or it should probably be a separate, non-automatic gradle invocation just running that task. I get the appeal of convenience, but correctness is what ultimately matters and this is working against how gradle tries to manage tasks IMO.

SUPERCILEX:
Would it feel more correct if we always ran the task and then just didn't do anything when the build fails? If yes, that's exactly what `onlyIf` is doing from the docs:
> The closure will be evaluated at task execution time, not during configuration.

If no, then I don't know how else you'd avoid committing when one of the publishing tasks fails.

ZacSweers:
I don't think a task should know anything about the build's failure. I would make the commit task explicitly depend on all publishing tasks (even if it doesn't use their outputs), that way it wouldn't run if one fails

SUPERCILEX:
Then how does a user only run certain publishing tasks? That solution would build an apk, a bundle, the listing, products, etc. What would be nice is if each publishing type could have its own commit, but that's not possible due the play API not allowing concurrent edits.

ZacSweers:
Hmm, maybe take a look at what this plugin does? https://github.com/vanniktech/gradle-maven-publish-plugin

It creates a repository on sonatype and then publishes artifacts to that repository, then finally closes+releases it at the end of the build. Maybe that's similar? I believe it uses a build service somewhere too

SUPERCILEX:
Took a cursory look and I don't think they have to deal with the "subsets of tasks can contribute to closing/committing" problem.

I think this works for now.

gmazzo:
> subsets of tasks can contribute to closing/committing

Exactly. I also check with @Vampire on Gradle's Slack and he also advised me to do something like this. I'm wondering if we should do a Feature Request on Gradle to extend the API to either have a `finalizedByIfSucceed` method or some kind of CC compatible `Provider<TaskOutcome>` that we can use with `onlyIf`.

Issue #1074, opened 2023-01-16
Add support for google-api-client 2.x.x


frmz:
### Problem description
Current version does't support google-api-client 2.x.x

### Potential solutions/workarounds
Upgrade android play store API to latest

### Additional context
Current failure message

`Caused by: java.lang.ExceptionInInitializerError
	at com.google.api.services.androidpublisher.AndroidPublisher$Builder.build(AndroidPublisher.java:11864)
	at com.github.triplet.gradle.androidpublisher.internal.AndroidPublisherKt.createPublisher(AndroidPublisher.kt:39)
	at com.github.triplet.gradle.androidpublisher.internal.DefaultPlayPublisher$Factory.create(DefaultPlayPublisher.kt:244)
	at com.github.triplet.gradle.androidpublisher.PlayPublisher$Companion.invoke(PlayPublisher.kt:109)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService$publisher$2.invoke(PlayApiService.kt:26)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService$publisher$2.invoke(PlayApiService.kt:21)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService.getPublisher(PlayApiService.kt)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService.getOrCreateEditId(PlayApiService.kt:79)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService.access$getOrCreateEditId(PlayApiService.kt:21)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService$edits$2.invoke(PlayApiService.kt:30)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService$edits$2.invoke(PlayApiService.kt:21)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at com.github.triplet.gradle.play.tasks.internal.PlayApiService.getEdits(PlayApiService.kt)
	at com.github.triplet.gradle.play.tasks.PublishBundle$BundleUploader.upload(PublishBundle.kt:103)
	... 30 more
Caused by: java.lang.IllegalStateException: You are currently running with version 2.1.1 of google-api-client. You need at least version 1.31.1 of google-api-client to run version 1.32.1 of the Google Play Android Developer API library.
[...]

PR #1090, opened 2023-06-26
Update play google-api-client 2.2.0 and android-publisher to 2.0.0

francescocervone:
Fixes https://github.com/Triple-T/gradle-play-publisher/issues/1074

I wanted to updated only the Google API Client, but the original issue was with the `android-publisher` being incompatible with Google API client 2.x.

I was not able to find a changelog for `android-publisher`. I just realized that last release was 2.0 by looking at the source: https://github.com/googleapis/google-api-java-client-services/tree/main/clients/google-api-services-androidpublisher/v3
Not sure how you handle these kinds of updates.

Google API client has an updated changelog instead: https://github.com/googleapis/google-api-java-client/releases.

I was able to integrate it in an existing project and it seems to work. Also, all unit tests pass.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -39,7 +39,7 @@ dependencyResolutionManagement {
             version("agp", "8.0.1")
             version("agp-tools", "31.0.1")
             version("android-publisher", "v3-rev20211021-1.32.1")
-            version("api-client", "1.32.2")
+            version("api-client", "2.2.0")
             version("http-client", "1.40.1")
             version("http-auth", "1.2.2")
             version("guava", "31.0.1-jre")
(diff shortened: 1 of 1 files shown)


francescocervone:
@SUPERCILEX where can I monitor the snapshot deploy? Both [PR](https://github.com/Triple-T/gradle-play-publisher/actions/runs/5380137603) and [master commit](https://github.com/Triple-T/gradle-play-publisher/actions/runs/5380485159) actions have skipped it because it looks like `version.txt` does not contain the snapshot version.

SUPERCILEX:
Oops, I just triggered a snapshot build. :)

francescocervone:
Everything works as expected on the snapshot 👍 Thanks!
Will you deploy a new release with this change?
We are unable to upgrade some other dependencies because of this issue.

SUPERCILEX:
Kicked off a build

Issue #763, opened 2020-01-31
Use last release notes when none are specified


vanniktech:
As a good default, I'd like to have the last release notes in case I didn't specify anything. This is better than having none at all. As for the selection process, the release notes that were used in the previously highest version code.

SUPERCILEX:
Hmmmm, I've thought of this before but haven't done anything because it's hard to know to know where the release notes should come from. You're suggesting picking from the highest version code, but that wouldn't work for places that automatically generate internal release notes for the internal/alpha channels. Would picking the release channel's previous changelog make sense?

vanniktech:
> Would picking the release channel's previous changelog make sense?

How about the channel we're publishing to? So if we publish `480` to `beta` and there's a changelog for `478` in `beta`, let's pick that one.

SUPERCILEX:
Dope, that's what I was trying to get at. I could do that. 👍

vanniktech:
Dope. Will be sure to test this out with the next release!

panpeter:
@SUPERCILEX at the moment the plugin seems to reuse the latest release notes only when it's a full rollout. Would it be possible to reuse release notes also when it's a staged rollout?

SUPERCILEX:
If there's an API to support that, sure.

PR #1105, opened 2023-11-01
Copy release notes from previous release also when it's a staged rollout

panpeter:
This PR updates the logic related to reusing release notes. 

Previously GPP was reusing the latest release notes only in case of a full rollout. Now GPP will reuse the latest release notes also when it's a staged rollout release. 

Discussed here: https://github.com/Triple-T/gradle-play-publisher/issues/763#issuecomment-1787267141

✅ Tests pass
✅ Checked with real project
✅ Documentation updated

### DIFF
--- a/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/TrackManager.kt
+++ b/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/TrackManager.kt
@@ -71,8 +71,6 @@ internal class DefaultTrackManager(
             createDefaultTrack(config)
         }
 
-        track.maybeCopyChangelogFromPreviousRelease(config.trackName)
-
         publisher.updateTrack(editId, track)
     }
 
@@ -118,8 +116,10 @@ internal class DefaultTrackManager(
                     }
                 }
             } else {
-                track.releases = track.releases +
-                        listOf(TrackRelease().mergeChanges(config.versionCodes, config.base))
+                val release = TrackRelease().mergeChanges(config.versionCodes, config.base).apply {
+                    maybeCopyChangelogFromPreviousRelease(config.trackName)
+                }
+                track.releases = track.releases + release
             }
         }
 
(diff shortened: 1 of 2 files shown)


play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/TrackManager.kt
```diff
@@ -71,8 +71,6 @@ internal class DefaultTrackManager(
             createDefaultTrack(config)
         }
 
-        track.maybeCopyChangelogFromPreviousRelease(config.trackName)
```
SUPERCILEX, line 74: Dumb question cuz I don't remember how anything works, but why was this removed?

SUPERCILEX:
Can you confirm the scenarios you've tested? I'm specifically interested in making sure release notes from past releases are still used if there are none in staging.

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/TrackManager.kt
panpeter: Instead of calling `Track.maybeCopyChangelogFromPreviousRelease` we now call `TrackRelease.maybeCopyChangelogFromPreviousRelease` in places where we create a `TrackRelease`. So this functionality is still there, just in different places.

panpeter:
> Can you confirm the scenarios you've tested? I'm specifically interested in making sure release notes from past releases are still used if there are none in staging.

I've primarily relied on `Track update uses existing release notes when no local ones are available` test. For example, if you comment out `maybeCopyChangelogFromPreviousRelease` the test does not pass. So we should be covered 😊

SUPERCILEX:
Alright sounds good, thanks! A snapshot build will be published in a bit. If you can make sure everything works with that build, I'd appreciate it.

panpeter:
Great, thanks!

> If you can make sure everything works with that build, I'd appreciate it.

Sure. We'll run a release on Monday. I'll let you know how it goes 👍

panpeter:
@SUPERCILEX , we've switched to the snapshot version and all works great. The release notes were copied from our previous releases 😊

SUPERCILEX:
Sweet, fired off a release!

PR #1114, opened 2023-12-13
Minor updates

Goooler:
Google Play now requires Apps published with target SDK 33 or higher, just bump the `compileSdk` and `targetSdk` in tests, see https://developer.android.com/google/play/requirements/target-sdk.

And we can kick these updates by enabling @renovate-bot.


.github/workflows/cid.yml
```diff
@@ -1,6 +1,12 @@
 name: CI/CD
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
```
SUPERCILEX, line 9: Undo. Or why?

.github/workflows/gradle-wrapper.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - 'gradlew'
+      - 'gradlew.bat'
+      - 'gradle/wrapper/'
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: gradle/wrapper-validation-action@v1
```
SUPERCILEX, line 15: Need to think about this more. Pretty sure I want this back in the `build` step so that it catches pushes too. Also that way it block publishing if validation fails.

renovate.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "config:base"
+    ]
+}
```
SUPERCILEX, line 6: Delete. I hate these dependency update bots because they spam you constantly. Why they don't default to a weekly/monthly combined update beats me.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
SUPERCILEX, line 3: Let's make this `-all` so the sources are available. Should be automatic by running `./gradlew wrapper --gradle-version 8.5`.

SUPERCILEX (changes requested):
Thanks for fixing this stuff!

gradle/wrapper/gradle-wrapper.properties
Goooler: Sources now could be attached by IDEA.

renovate.json
Goooler: We can configure it to update weekly.

.github/workflows/gradle-wrapper.yaml
Goooler: This action is only useful when Gradle wrapper is changed, we don't need to run it for every check.

.github/workflows/cid.yml
Goooler: We can let CI run on master, PRs, and manually triggering by `workflow_dispatch`, no need double checks in every PR.

SUPERCILEX: Nah, there's no need to run manually because builds should be idempotent. And I want it to run on every branch because I don't use PRs for my own changes.

Revert.

gradle/wrapper/gradle-wrapper.properties
SUPERCILEX: K, then remove this stuff everywhere https://github.com/Goooler/gradle-play-publisher/blob/59b0a92a6a58318cd0cf54294328451b2ca7720a/testapp/build.gradle.kts#L39

And run `./gradlew wrapper --gradle-version 8.5` both top-level and in testapp.

.github/workflows/cid.yml
Goooler: 2abd688491737450ed91a851264a4182d203b64d

.github/workflows/gradle-wrapper.yaml
SUPERCILEX: Nah, revert. I want the proof authenticity to be in the same build that publishes stuff.

Goooler: cf4c91736565269d927507343ac1ebcfac460caf

gradle/wrapper/gradle-wrapper.properties
Goooler: 880950acffd7931c9203960c64b02ca099a1ae68

renovate.json
SUPERCILEX: Hmmm, for now let's not do this. If I was actively developing the project, I'd say yes but it's been in maintenance mode for a few years now and I'd rather just leave things alone as long as they work.

Revert and then I'll merge! Thanks for doing this!

Goooler: 070470784aad5f1cd6231e7cfddb340fdc88836d

SUPERCILEX:
Thanks again!

PR #1117, opened 2023-12-23
Shade third-party dependencies in the plugin classpath

### LINKED ISSUE #1091: Consider shading Google API client dependencies
### Problem description

I have Google API client dependencies on the build classpath from [`com.google.firebase.testlab:testlab-gradle-plugin`](https://maven.google.com/web/index.html#com.google.firebase.testlab:testlab-gradle-plugin:0.0.1-alpha03). That plugin needs version 1.x of `google-api-client`, and this plugin wants version 2.x, which are not compatible:

```
Exception java.lang.IllegalStateException: You are currently running with version 2.2.0 of google-api-client. You need at least version 1.15 of google-api-client to run version 1.22.0 of the Cloud Testing API library.
```

### Potential solutions/workarounds

Ideally, both plugins would shade their dependencies, for example by using the [`shadow` plugin](https://github.com/johnrengelman/shadow); this is because Gradle doesn't isolate classpaths between plugins, so it ends up resolving the larger of the two library versions per its normal dependency resolution rules. 

I will also open a report on Google's issue tracker.

tadfisher:
Fixes #1091 

At least one other Android-related plugin also puts Google API client libraries on the classpath with conflicting versions; see the linked issue. 

- Apply the [shadow plugin](https://imperceptiblethoughts.com/shadow/introduction/) to `android-publisher` and `play-publisher`.
- Add a `shadowImplementation` configuration, for which dependency classes will be relocated to a package under `com.github.triplet.gradle.shaded`. Move all non-project dependencies from `implementation` to `shadowImplementation`.
- Publishing the shaded jar is configured automatically by `com.gradle.plugin-publish`. As `android-publisher` is not a plugin itself and doesn't apply that plugin in its build, apply `PublishTaskShadowAction` as `plugin-publish` does.
- `testapp`: Bump `compileSdk` and `targetSdk` to 34 so it builds with AGP 8.2.0.

Tested locally by publishing to `mavenLocal` and using it in another project, by running `check` tasks, and by running `testapp:publishBundle` (which fails due to an auth token issue, but otherwise applies the plugin fine).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,5 @@
+import com.github.jengelman.gradle.plugins.shadow.ShadowPlugin
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import io.github.gradlenexus.publishplugin.CloseNexusStagingRepository
 import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import java.time.Duration
@@ -15,6 +17,7 @@ plugins {
     // Needed to support publishing all modules atomically
     alias(libs.plugins.gradlePublish) apply false
     alias(libs.plugins.nexusPublish)
+    alias(libs.plugins.shadow) apply false
 }
 
 buildScan {
@@ -81,6 +84,25 @@ allprojects {
 
             useInMemoryPgpKeys(System.getenv("SIGNING_KEY"), System.getenv("SIGNING_PASSWORD"))
         }
+    }
+
+    plugins.withType<ShadowPlugin> {
+        val shadowImplementation by configurations.creating
+        configurations["compileOnly"].extendsFrom(shadowImplementation)
+        configurations["testImplementation"].extendsFrom(shadowImplementation)
+
+        tasks.withType<ShadowJar> {
+            archiveClassifier.set("")
+            configurations = listOf(shadowImplementation)
+            isEnableRelocation = true
+            relocationPrefix = "com.github.triplet.gradle.shaded"
+        }
+    }
+
+    // Needed to preserve JAR hash for testapp build
+    tasks.withType<AbstractArchiveTask> {
+        isPreserveFileTimestamps = false
+        isReproducibleFileOrder = true
     }
 
     tasks.withType<Test> {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,9 @@
 import com.github.jengelman.gradle.plugins.shadow.ShadowPlugin
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import io.github.gradlenexus.publishplugin.CloseNexusStagingRepository
-import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
 import java.time.Duration
+import org.jetbrains.kotlin.gradle.dsl.KotlinProjectExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
 
 buildscript {
     dependencies {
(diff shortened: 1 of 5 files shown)


SUPERCILEX:
Can you send me a PR with all the non shadow stuff? So `plugins.withType` and the testapp sdk.

I don't like this because it defeats the purpose of using a dependency manager. For example, before this change people could include versions of libraries we use in their builds as a way to update us without having to get a new version of GPP. Similarly, I don't know if this is actually how the shadow plugin works, but I'm guessing it won't show any of our dependencies in the POM which means security scanning tools also won't work. We're basically opting out of the ecosystem. Maybe that's fine? I need to think about it a bit more.

Have you filed a ticket with Firebase support? They really should be the ones to upgrade.

tadfisher:
> Can you send me a PR with all the non shadow stuff? So plugins.withType and the testapp sdk.

Will do.

> I don't like this because it defeats the purpose of using a dependency manager. For example, before this change people could include versions of libraries we use in their builds as a way to update us without having to get a new version of GPP.

I completely agree. The problem is on the Gradle side, though; plugins run without classpath isolation, so shading dependencies is the recommended approach by upstream (and why they built support for the Shadow plugin in `plugin-publish`). Otherwise you're destined to get ABI conflicts with *any* third-party dependency declared in the POM, because Gradle will resolve the highest version to include in the global, non-isolated plugin classpath.

There is another approach you could take, [which is done by the Android Lint plugin](https://android.googlesource.com/platform/tools/base/+/studio-master-dev/lint/libs/lint-gradle-api/README.md): split the plugin into API and implementation artifacts, and have the API code construct a semi-isolated "embedded" classpath loader which loads runtime dependencies, making them available via reflection. This is less simple than shading, and arguably should be performed by Gradle itself, but at least it preserves the dependency information in the POM. You're already halfway there, having a separate `android-publisher` artifact with the majority of third-party dependencies.

Reading up on this: as you're using the Worker API, we could configure workers to use [classLoaderIsolation](https://docs.gradle.org/current/userguide/worker_api.html#changing_the_isolation_mode) and include Google API dependencies in a separate dependency configuration. This might be the cleanest way to avoid classpath conflicts, and it seems at least to be the most supported method.

> Similarly, I don't know if this is actually how the shadow plugin works, but I'm guessing it won't show any of our dependencies in the POM which means security scanning tools also won't work. We're basically opting out of the ecosystem.

Yes, as-is this PR will remove the dependency information from the POM/Gradle-module metadata.

> Have you filed a ticket with Firebase support? They really should be the ones to upgrade.

[Ticket](https://issuetracker.google.com/issues/289749733) was filed in July; no response just yet, though it was assigned.

I think I will investigate Worker classpath isolation, unless you have objections. There is probably a bit of API leakage into the plugin currently.

SUPERCILEX:
> Will do.

See below.

> so shading dependencies is the recommended approach by upstream

Alright, that works for me! I want this to be an isolated change so I can easily revert it if the pitchforks show up on the other side of this debate, but separate commits will suffice. So just pull out the unrelated changes in a separate commit and we're good.

> [Ticket](https://issuetracker.google.com/issues/289749733) was filed in July; no response just yet, though it was assigned.

Classic google lol.

> I think I will investigate Worker classpath isolation, unless you have objections.

If you'd like to, go for it! It does sound like the right thing to do. That said, it also sounds like a lot of work :sweat_smile:, so I'm ok going with the shadow approach until someone complains about it.

tadfisher:
I investigated using classLoaderIsolation() for the worker tasks, similar to what SQLDelight is doing here: https://github.com/cashapp/sqldelight/blob/a188d6afb52afce9a736047206ed08405bd8cc3d/sqldelight-gradle-plugin/src/main/kotlin/app/cash/sqldelight/gradle/SqlDelightWorkerTask.kt#L25

It is sort of hacky, because `PlayApiService` is shared among all workers, but that's the thing we want to use classpath isolation; basically, the service loads a PlayPublisher implementation using the ClassLoader of the first worker thread that instantiates the lazy `publisher` instance, which doesn't make a whole lot of sense. I tried using a private classloader within `PlayApiService`, but Gradle doesn't let you do this outside of a worker thread (they detect this during configuration resolution).

So I cleaned up this change instead.

SUPERCILEX:
> because PlayApiService is shared among all workers

Yeah, trying to accumulate changes to an edit in parallel and then commit the edit at the end has been the bane of this project's existence lol.

build.gradle.kts
```diff
@@ -58,31 +62,56 @@ allprojects {
...
             useInMemoryPgpKeys(System.getenv("SIGNING_KEY"), System.getenv("SIGNING_PASSWORD"))
         }
     }
 
+    plugins.withType<ShadowPlugin> {
+        val shadowImplementation by configurations.creating
+        configurations["compileOnly"].extendsFrom(shadowImplementation)
+        configurations["testImplementation"].extendsFrom(shadowImplementation)
+
+        tasks.withType<ShadowJar> {
```
Goooler, line 101:
```suggestion
        tasks.withType<ShadowJar>().configureEach {
```

build.gradle.kts
```diff
@@ -58,31 +62,56 @@ allprojects {
...
+            archiveClassifier.set("")
+            configurations = listOf(shadowImplementation)
+            isEnableRelocation = true
+            relocationPrefix = "com.github.triplet.gradle.shaded"
+        }
+    }
+
+    // Needed to preserve JAR hash for testapp build
+    tasks.withType<AbstractArchiveTask> {
```
Goooler, line 110:
```suggestion
    tasks.withType<AbstractArchiveTask>().configureEach {
```

SUPERCILEX:
Feel free to reopen once you're ready.

PR #1123, opened 2024-03-19
Use new github action output command

StefMa:
I just came across the old usage of defining output.
I changed it.
Honestly, I'm not sure if it works 😅 
But I also have never seen such an setup anyways.. So 🤷 

More info about the deprecation:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #1124, opened 2024-03-21
Use new github actions output syntax

StefMa:
Better solution than #1123 🙃 

This time I also tested it on my fork.
See a non-snapshot version run:
https://github.com/StefMa/gradle-play-publisher/actions/runs/8371003805

vs a snapshot version run:
https://github.com/StefMa/gradle-play-publisher/actions/runs/8371010999
This one fails because I can't publish snapshots on the fork.
However, the idea is that the job `deploy_snapshot` runs; and this is what it does.

### DIFF
--- a/.github/workflows/cid.yml
+++ b/.github/workflows/cid.yml
@@ -23,7 +23,9 @@ jobs:
         uses: gradle/wrapper-validation-action@v1
       - name: Setup environment
         id: setup
-        run: ./gradlew configureGithubActions
+        run: |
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
Lol, I like that you peaced out of Gradle. Also duh, we were trying to use bash syntax in a println, obviously won't work. My bad for not noticing.

SUPERCILEX:
Nice, it worked! https://github.com/Triple-T/gradle-play-publisher/actions/runs/8371170014

Issue #1067, opened 2022-10-10
Missing support for Subscriptions API


vlad-roid:
gradle-play-publisher 3.7 only supports legacy subscription products created last year or earlier, not new ones from this year which are not returned at all by the In-App Products API and have a very different structure
![image](https://user-images.githubusercontent.com/26101312/194789804-50bb71da-1c2c-4da7-9797-17cb7c28f034.png)

Please add support for the Subscriptions API, otherwise there's no way to use the library for those as they're not even generating by `bootstrapListing --products`.

vlad-roid:
@SUPERCILEX sorry do you know if this is coming soon? InApp Products API is pretty much deprecated, there's no turning back to it unfortunately and `bootstrapListing --products` isn't returning the subscription products.

SUPERCILEX:
That's a bummer, but I don't actively maintain this plugin anymore. That said, I'm happy to review PRs.

vlad-roid:
Got it, thank you. Hopefully you'll see a PR from me for this soon then ;) Unsure how much effort the migration to the new API would be, Google I guess is known for overcomplicating things.

PR #1126, opened 2024-05-13
Support for Subscriptions API (issue 1067)

kb0:
PR that fixes https://github.com/Triple-T/gradle-play-publisher/issues/1067

* loads subscriptions from Google Play as json (like products) into 'play/subscriptions' using monetization.subscriptions/list
`gradlew bootstrapListing --subscriptions`
* publish subscriptions using monetization.subscriptions/create and monetization.subscriptions/patch
`gradlew publishSubscriptions`


play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
```diff
@@ -212,12 +215,67 @@ internal class DefaultPlayPublisher(
...
+                }
+                .execute()
+    }
+
+    override fun updateInAppSubscription(subscriptionFile: File): UpdateSubscriptionResponse {
+        val subscription = readSubscriptionFile(subscriptionFile)
+        try {
+            publisher.monetization().subscriptions().patch(subscription.packageName, subscription.productId, subscription)
+                    .apply {
+                        regionsVersionVersion = SUBSCRIPTIONS_REGIONS_VERSION
```
SUPERCILEX, line 252: Thoughts on sticking this in the subscriptions file name? Otherwise I don't know how we'll ever be able to upgrade people without a breaking release.

The idea is to have `insertInAppSubscription` and `updateInAppSubscription` take in a `regionsVersion` that the bootstrap task and publish subscriptions task serialize and deserialize into the file name (maybe `product_id.regionVer.json`?). Basically just use the file name but move validation etc outside of the api module which is supposed to be dumb.

Using the file name is bad because it's an implicit place to store untyped data, but good because then each subscription can be on its own version. Of course the more complicated but correct way to do this is with a separate metadata file but that seems way too annoying.

play/android-publisher/src/testFixtures/kotlin/com/github/triplet/gradle/androidpublisher/FakePlayPublisher.kt
```diff
@@ -35,6 +35,17 @@ abstract class FakePlayPublisher : PlayPublisher {
...
+    override fun getInAppSubscriptions(): List<GppSubscription> =
+            throw IllegalStateException("Test wasn't expecting this method to be called.")
+
+    override fun insertInAppSubscription(subscriptionFile: File): Unit =
+            throw IllegalStateException("Test wasn't expecting this method to be called.")
+
+    override fun updateInAppSubscription(subscriptionFile: File): UpdateSubscriptionResponse =
+            throw IllegalStateException("Test wasn't expecting this method to be called.")
+
+    // TODO add subscriptions methods
```
SUPERCILEX, line 47: Remove?

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/CliOptions.kt
```diff
@@ -265,6 +269,12 @@ internal interface BootstrapOptions {
                 field = value
             }
 
+        override var downloadSubscriptions = true
```
SUPERCILEX, line 272: Should `downloadProducts` (line 266 above) default to false? Like does it even make sense to have both?

SUPERCILEX (changes requested):

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/CliOptions.kt
kb0: No. We should have both "products" and "subscriptions" to true. "Products" (inappproducts.*) still used for in-app purchases (consumable and non-consumable), and "subscriptions" (monetization.subscriptions.*) used only for subscriptions.

play/android-publisher/src/testFixtures/kotlin/com/github/triplet/gradle/androidpublisher/FakePlayPublisher.kt
kb0: Yes, sorry for forgotten "TODO" :)

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
kb0: May be it's better to use regions version as plugin field, i.e `play { subscriptionsRegionsVersion = '2022/02' }`?
I cannot find any data on RegionsVersion in 'monetization.subscriptions.list' - so after bootstrapping there is no information about current regions version.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/CliOptions.kt
SUPERCILEX: Sounds good, thanks for the explanation!

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/DefaultPlayPublisher.kt
SUPERCILEX: Ah, well that sucks. I don't like putting the config in `play {}` because if you don't specify the `subscriptionsRegionsVersion`, then that's the same as having a hardcoded default. Unless we default to null and error out if you haven't specified it when bootstrapping/publishing? I also don't like the fact that it forces you to upgrade everything at once if you want to go to a newer version, but maybe that's not a big deal?

I dunno, tradeoffs tradeoffs tradeoffs...

kb0: As I understand, this value changes quite rarely - so "2022/02" may be better than "null".
The same reason for single value per instance, instead of value per each subscription.

If some one needs, for ex., to update one set of subscriptions with "2022/02", other with "2022/01" - it's possible to do some "dirty hacks" like - create own tasks `mySubscription01`, `mySubscription02` - copy jsons from somewhere into ./play/subscriptions, set play.subscriptionsRegionsVersion, run publishSubscription 🤦‍♀️

SUPERCILEX: > so "2022/02" may be better than "null"

Well but then there's no point in having that `play {}` option because if we updated the default from "2022/02" to 2024/whatever, then all the people that forgot to set the `play {}` are silently broken.

So our options are between having a `play {}` option that errors if unspecified when using subscriptions stuff or building the version into the file name. That means the decision comes down to whether or not we want to allow subscriptions to use independent region versions. With that in mind, do you still have a preference? Otherwise I prefer the file name approach for its flexibility.

kb0: Ok. But may be it's better to put versions in seprate files, because values contains "/", like:
`product_id.metadata.json`:
{ "regionsVersion": "2022/02" }

SUPERCILEX: Or replace with a dash. But yeah, maybe a metadata file is just better. Are you ok with doing that?

kb0: Yes

SUPERCILEX: Cool!

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/Bootstrap.kt
```diff
@@ -247,7 +249,11 @@ internal abstract class Bootstrap @Inject constructor(
 
             val subscriptions = apiService.publisher.getInAppSubscriptions()
             for (subscription in subscriptions) {
-                parameters.dir.get().file("${subscription.productId}.json").write(subscription.json)
+                parameters.dir.get().file("${subscription.productId}.json")
+                        .write(subscription.json)
+                parameters.dir.get().file("${subscription.productId}.$SUBSCRIPTION_METADATA_SUFFIX.json")
+                        .write(GsonFactory.getDefaultInstance().toString(
+                                mapOf("regionsVersion" to "")))
```
SUPERCILEX, line 256: Here we do want a default, so let's add the 2022/02 constant.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishSubscriptions.kt
```diff
@@ -49,31 +49,52 @@ internal abstract class PublishSubscriptions @Inject constructor(
 
     @TaskAction
     fun publishSubscriptions(changes: InputChanges) {
+
```
SUPERCILEX, line 52: Delete the new line

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishSubscriptions.kt
```diff
@@ -49,31 +49,52 @@ internal abstract class PublishSubscriptions @Inject constructor(
...
         changes.getFileChanges(subscriptionsDir)
                 .filterNot { it.changeType == ChangeType.REMOVED }
                 .filter { it.fileType == FileType.FILE }
+                .map {
+                    if (it.file.name.endsWith(".$SUBSCRIPTION_METADATA_SUFFIX.json")) {
+                        it.file.parentFile.resolve(
+                                it.file.name.replace(
+                                        "\\.$SUBSCRIPTION_METADATA_SUFFIX\\.json$".toRegex(), ".json"))
```
SUPERCILEX, line 60: Pull `".$SUBSCRIPTION_METADATA_SUFFIX.json"` out into a local variable and make this `it.file.name.dropLast(metadataFileSuffix.length) + ".json"`.

Also add a comment explaining that we need these potential duplicates so that if someone makes an incremental change to just the metadata file, we'll still notice that and try to publish the updated subscription.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishSubscriptions.kt
```diff
@@ -49,31 +49,52 @@ internal abstract class PublishSubscriptions @Inject constructor(
...
+            val metadata = subscriptionFile.parentFile
+                    .resolve("${subscription["productId"]}.$SUBSCRIPTION_METADATA_SUFFIX.json").inputStream().use {
+                        GsonFactory.getDefaultInstance().createJsonParser(it).parse(Map::class.java)
+                    }
 
             println("Uploading ${subscription["productId"]}")
-            val response = apiService.publisher.updateInAppSubscription(subscriptionFile)
-            if (response.needsCreating) apiService.publisher.insertInAppSubscription(subscriptionFile)
+            val response = apiService.publisher.updateInAppSubscription(subscriptionFile, metadata["regionsVersion"] as String)
+            if (response.needsCreating) apiService.publisher.insertInAppSubscription(subscriptionFile, metadata["regionsVersion"] as String)
```
SUPERCILEX, line 88: Here and for serialization, could we have a class instead? Seems much nicer than this untyped stuff.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishSubscriptions.kt
```diff
@@ -49,31 +49,52 @@ internal abstract class PublishSubscriptions @Inject constructor(
...
         }
 
         interface Params : PlayPublishingParams {
             val target: RegularFileProperty
         }
     }
+
+    companion object {
+        /** The suggested environment variable name in which to store credentials. */
```
SUPERCILEX, line 97: Fix: `The file name suffix for subscription metadata files.`

README.md
```diff
@@ -596,6 +596,13 @@ Run `./gradlew publishSubscriptions`.
 Manually setting up in-app subscriptions files is not recommended. [Bootstrap them instead](#quickstart)
 with `./gradlew bootstrapListing --subscriptions`.
 
+Each subscription bootstraps with additional metadata-file (`subscriptions/<subscription product id>.metadata.json`)
+which should contains `regionsVersion` code (https://developers.google.com/android-publisher/api-ref/rest/v3/RegionsVersion, latest is `2022/02`).
+Using this field will ensure that creating and updating the resource with an older region's version
+and set of regional prices and currencies will succeed even though a new version is available.
+Currently, there is no way to get this value from the Google Play Developer API, so you should
+manually set regions version for each subscription metadata file.
```
SUPERCILEX, line 604: TODO(alex): look into tweaking the wording.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/PublishSubscriptions.kt
```diff
@@ -49,31 +49,52 @@ internal abstract class PublishSubscriptions @Inject constructor(
...
 
     abstract class Uploader : PlayWorkerBase<Uploader.Params>() {
         override fun execute() {
             val subscriptionFile = parameters.target.get().asFile
+
             val subscription = subscriptionFile.inputStream().use {
                 GsonFactory.getDefaultInstance().createJsonParser(it).parse(Map::class.java)
             }
+            val metadata = subscriptionFile.parentFile
+                    .resolve("${subscription["productId"]}.$SUBSCRIPTION_METADATA_SUFFIX.json").inputStream().use {
```
SUPERCILEX, line 82: This needs to be `subscriptionFile.name.dropLast("json".length) + "$SUBSCRIPTION_METADATA_SUFFIX.json"` so people aren't forced to name their files after the product id.

SUPERCILEX (changes requested):

SUPERCILEX:
This is great, thanks for the PR!

vlad-roid:
thanks so much guys!

PR #1133, opened 2024-06-24
Github release with autom. generated changelog

hannesa2:
When you set a new tag, now a release will be created here https://github.com/Triple-T/gradle-play-publisher/releases and the changelog will be automated generated by containing pull requests.
You can make `xxx-alphax` and `xxx-betax` release as well. They will be crated as pre-release
image

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,25 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+    authors:
+      - someuser
+  categories:
+    - title: Breaking Changes 🛠
+      labels:
+        - breaking-change
+    - title: Exciting New Features 🎉
+      labels:
+        - enhancement
+    - title: Dependencies
+      labels:
+        - dependencies
+    - title: Espresso test
+      labels:
+        - Espresso
+    - title: Pipeline
+      labels:
+        - pipeline
+    - title: Other Changes
+      labels:
+        - "*"
--- a/.github/workflows/cid.yml
+++ b/.github/workflows/cid.yml
@@ -98,3 +98,34 @@ jobs:
           SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
           SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
+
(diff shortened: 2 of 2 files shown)


SUPERCILEX:
As much as I hate rejecting PRs, I can't accept this for a few reasons:
- Security: this uses several actions from different people that I'd need to look into to decide if I trust.
- I don't follow a PR-only model: sometimes (as in the last release) fixes and features appear in commits I push directly to master. As far as I understand, those wouldn't show up in the release notes.
- Publishing delay: this is a lame excuse, but since I don't use any of this stuff myself anymore, I like to wait a bit before sending out the release notification so any early adopters would hopefully catch bugs for me. Kinda dumb, but whatever.

I do appreciate you trying to help me out though, thanks!

hannesa2:
### A: Security
Speaking about security, 
> I'd need to look into to decide if I trust.

Same to me, I've to trust this plugin as well and need the most transparent way for releasing. Best is automated.
 
I'm on the other side (I'm a consumer of this nice plugin), and when I read 
> Fixed a crash bootstraping when no subscriptions exist.
image

I would love to have in release note a link, where I can immediate see related code changes

### B: PR model
> I don't follow a PR-only model:

Yes, this is an issue, or a feature. 
As a consumer, I would like to see all changes in release.
As a producer, I could use this to hide changes any any reasons in release

### C: Delay
> Publishing delay

For this the alpha/beta-way (prerelease) is the solution

SUPERCILEX:
> Same to me, I've to trust this plugin as well and need the most transparent way for releasing. Best is automated.

So I looked into it and I'd trust the softprops one.

> I would love to have in release note a link, where I can immediate see related code changes

Hmmm, fair enough.

> As a producer, I could use this to hide changes on any reasons in release.

There's not much thought going into this on my part, it's just laziness. Also, it's not that hard to just pull up a diff of the tags yourself to see everything.

> For this the alpha/beta-way (prerelease) is the solution

Good point, I'd consider doing that.

---

Ok, I if this was trimmed down to just the softprops action with auto generated release notes I'd accept it. Though it's unclear to me what the generated release notes will look like. Basically I'd be willing to try it for one release to see what happens.

hannesa2:
Here https://github.com/gitx/gitx/releases you can see a sample, how does the changelog looks like

SUPERCILEX:
Nice! Yeah that looks good.

Issue #1088, opened 2023-05-04
Support for form factor


ibatica:
### Problem description
Currenlty it is possible to upload only for phone/tablet form, "automotive" and "wear" forms are not support.
### Potential solutions/workarounds
In order to support form factor by Android Publishing API track should be defined as wear:production" or "wear:beta", "automotive:internal"
This doesn't work with current implementation.
### Additional context
Android Publishing API track and form factor https://developers.google.com/android-publisher/tracks

JBassett:
Are there any workarounds other than manually deploying?

markchristopherng:
Hi,  we also require this for our wear app where Google requires all wear apps to be on a seperate track by end of August 2023 

https://support.google.com/googleplay/android-developer/answer/13295490?hl=en

sorinirimies:
We have a wear app and since Google has started enforcing form factor publishing we can't use properly this plugin.
A form factor option would really be great.

crysxd:
Is there any workaround? We are kinda slowly approaching the deadline. I assume we need to enable the form factor track and then publish the wear app manually?

adrianolc:
We're running into the same problem now. Is there any update on when it will be supported or any workaround to solve the issue?

SUPERCILEX:
I'm not involved with the project anymore, but I still do PR reviews if people want to fix this.

damiafuentes:
We're facing the same issue.

crysxd:
I looked at this problem again as it has to be solved soon and now it seems to all work for me as expected. 

What I did:
- Create the Wear track in Google Play, this was offered to me in a banner in the Release overview in the Google Play Console
- Update my wear gradle config
```
play {
    serviceAccountCredentials.set(project.rootProject.ext.play_credentials)
    track.set("wear:" + project.findProperty("play.track"))
    defaultToAppBundles.set(true)
    if (project.rootProject.ext.play_release_fraction >= 1d) {
        releaseStatus.set(com.github.triplet.gradle.androidpublisher.ReleaseStatus.COMPLETED)
    } else {
        releaseStatus.set(com.github.triplet.gradle.androidpublisher.ReleaseStatus.IN_PROGRESS)
    }
    updatePriority.set(project.rootProject.ext.play_update_priority)
    userFraction.set(project.rootProject.ext.play_release_fraction)
    releaseName.set(project.rootProject.ext.version_name)
}
```
- For reference, my phone gradle config:
```
play {
[...]

PR #1134, opened 2024-06-26
Handle release track names containing a colon

chimbori:
Handle release track names containing a colon (e.g. wear:alpha) gracefully without crashing.

https://github.com/Triple-T/gradle-play-publisher/issues/1088

For releases named `formfactor:track`, try `formfactor-track.txt` and `track.txt` instead of invalid filename `formfactor:track.txt`

https://github.com/chimbori/gradle-play-publisher/commit/cb5729afa399e1980231af67cf6670647d479915

### DIFF
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PublishArtifactWorkerBase.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PublishArtifactWorkerBase.kt
@@ -25,7 +25,6 @@ internal abstract class PublishArtifactWorkerBase<T : PublishArtifactWorkerBase.
             val dir = parameters.consoleNamesDir.get()
             val file = dir.file("${track.replace(':', '-')}.txt").asFile.orNull()
                     ?: dir.file("${track.substringAfter(':')}.txt").asFile.orNull()
-                    ?: dir.file("$track.txt").asFile.orNull()
                     ?: dir.file(RELEASE_NAMES_DEFAULT_NAME).asFile.orNull()
             file?.readProcessed()?.lines()?.firstOrNull()
         } else {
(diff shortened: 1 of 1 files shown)


play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/workers/PublishArtifactWorkerBase.kt
```diff
@@ -22,10 +22,11 @@ internal abstract class PublishArtifactWorkerBase<T : PublishArtifactWorkerBase.
...
             config.releaseName
         } else if (parameters.consoleNamesDir.isPresent) {
-            val dir = parameters.consoleNamesDir.get().asFile
-            val file = File(dir, "$track.txt").orNull()
-                    ?: File(dir, RELEASE_NAMES_DEFAULT_NAME).orNull()
-
+            val dir = parameters.consoleNamesDir.get()
+            val file = dir.file("${track.replace(':', '-')}.txt").asFile.orNull()
+                    ?: dir.file("${track.substringAfter(':')}.txt").asFile.orNull()
+                    ?: dir.file("$track.txt").asFile.orNull()
```
SUPERCILEX, line 28: Shouldn't this be removed entirely? If you don't have a file named `formfactor-track.txt` or `track.txt`, it's still going to crash with this right as it'll try `formfactor:track.txt`. Pretty sure `track.substringAfter(':')` covers `track` as if there's no colon it'll do nothing.

chimbori: Good catch, I’m not sure whether simply calling `.file("file:with:colon.txt")` will crash or return a null instead, but like you said, since `.substringAfter(':')` covers this, we can just remove this line.

PR #1145, opened 2024-09-04
Fix the project isolation issues

### LINKED ISSUE #1144: Plugin not compatible with project isolation
### Describe the bug

Gradle has a new feature called [project isolation](https://docs.gradle.org/current/userguide/isolated_projects.html) which will allow modules to be configured in parallel. To allow this only certain things can be accessed on a `Project` from another module.

Currently this plugin has 2 violations:
image

### How To Reproduce
Simply apply the plugin and run `./gradlew -Dorg.gradle.unsafe.isolated-projects=true`

#### Versions
- Gradle Play Publisher: 3.10.1
- Gradle Wrapper:  8.10
- Android Gradle Plugin: 8.6.0

#### Tasks executed
n/a

### Expected behavior
This plugin should be compatible with project isolation.

ansman:
With the new project isolation feature in Gradle there are restrictions on what you can access on a `Project` instance for another module.

This fixes #1144


common/validation/src/main/kotlin/com/github/triplet/gradle/common/validation/Validation.kt
```diff
@@ -10,15 +10,15 @@ import org.gradle.api.logging.Logger
...
 /**
  * Validates required dependencies. If GPP can't run in the current context, an error will be
  * thrown.
  */
 fun Project.validateRuntime() {
-    rootProject.plugins.apply(RuntimeValidationPlugin::class)
-
-    project.plugins.withType<AppPlugin> {
+    GradleRuntimeValidator(GradleVersion.current(), MIN_GRADLE_VERSION).validate()
+    plugins.withType<AppPlugin> {
```
ansman, line 21: I see no reason why this code had to live in a plugin nor why it had to only be applied to the root project. Either it will fail and then the build will stop or it will pass and then pass for every module this plugin is applied to.

SUPERCILEX: I think there were classpath issues back in the day, but probs doesn't matter anymore.

ansman: Oh I see, but the Gradle version should be consistent across all modules?

SUPERCILEX: Yeah, I think this is fine. Did you see the build failure? That's what I'm waiting on.

ansman: I did, it has been fixed now.

PR #1148, opened 2024-11-05
Add Application Default Credentials & Service Account impersonation

a-mackay:
## Context

Usage of Service Account keys is [generally discouraged](https://cloud.google.com/iam/docs/best-practices-service-accounts#service-account-keys), with service account [impersonation](https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use) being a recommended alternative.

## Changes

Allow [Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) to be used, and support Service Account impersonation with it. This could be used as an alternative to the existing Service Account key authentication. Since it is an alternative method, it should not affect backwards compatibility.

One point of discussion is the default values used to build `ImpersonatedCredentials` in `AndroidPublisher.kt`. We think these are suitable defaults - hardcoded values for the token lifetime was taken from GCP documentation and seems adequate given that tokens are autogenerated/refreshed by the underlying SDK as HTTP requests are constructed.

### DIFF
--- a/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
+++ b/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
@@ -14,6 +14,7 @@ import com.google.api.services.androidpublisher.AndroidPublisher
 import com.google.api.services.androidpublisher.AndroidPublisherScopes
 import com.google.auth.http.HttpCredentialsAdapter
 import com.google.auth.oauth2.GoogleCredentials
+import com.google.auth.oauth2.ImpersonatedCredentials
 import org.apache.http.auth.AuthScope
 import org.apache.http.auth.UsernamePasswordCredentials
 import org.apache.http.impl.client.BasicCredentialsProvider
@@ -39,6 +40,30 @@ internal fun createPublisher(credentials: InputStream): AndroidPublisher {
             transport,
             GsonFactory.getDefaultInstance(),
             AndroidPublisherAdapter(credential)
+    ).setApplicationName(PLUGIN_NAME).build()
+}
+
+internal fun createPublisher(impersonateServiceAccount: String?): AndroidPublisher {
+    val transport = buildTransport()
+
+    val appDefaultCreds = GoogleCredentials.getApplicationDefault()
+
+    val credential = if (impersonateServiceAccount != null) {
+        ImpersonatedCredentials.newBuilder()
+                .setSourceCredentials(appDefaultCreds)
+                .setTargetPrincipal(impersonateServiceAccount)
+                .setScopes(listOf(AndroidPublisherScopes.ANDROIDPUBLISHER))
+                .setLifetime(300)
+                .setDelegates(null)
+                .build()
+    } else {
+        appDefaultCreds
+    }
+
+    return AndroidPublisher.Builder(
+            transport,
+            GsonFactory.getDefaultInstance(),
+            AndroidPublisherAdapter(credential as GoogleCredentials)
     ).setApplicationName(PLUGIN_NAME).build()
 }
 
--- a/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
+++ b/play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
@@ -53,7 +53,6 @@ internal fun createPublisher(impersonateServiceAccount: String?): AndroidPublish
                 .setSourceCredentials(appDefaultCreds)
                 .setTargetPrincipal(impersonateServiceAccount)
                 .setScopes(listOf(AndroidPublisherScopes.ANDROIDPUBLISHER))
-                .setLifetime(300)
                 .setDelegates(null)
                 .build()
     } else {
--- a/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
+++ b/play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
@@ -17,7 +17,6 @@ import org.gradle.api.services.BuildServiceParameters
 import org.gradle.tooling.events.FinishEvent
 import org.gradle.tooling.events.OperationCompletionListener
 import org.gradle.tooling.events.task.TaskFailureResult
-import org.gradle.tooling.events.task.TaskFinishEvent
 import java.io.ByteArrayInputStream
 import java.io.InputStream
 import javax.inject.Inject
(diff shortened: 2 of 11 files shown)


SUPERCILEX:
Did you test this locally?

a-mackay:
Yep, we built the plugin locally, and published our Android app successfully. It used my personal Google account set up via the `gcloud` CLI, and impersonated our publisher Service Account.

pauldthomson:
Just to add, we contemplated some more tests around this but the only "logic" added resulted in a different instantiation of a GoogleCredentials, so it meant several levels of mocking to assert on a type and the logic is fairly simple so it didn't seem like the cost of the test added much value. 
The actual logic of using ADC and impersonation lives in the Google SDK anyway..

SUPERCILEX:
Good enough for me, thanks! I'll try to review this before the end of the weekend.

SUPERCILEX:
Can you give me edit access to the PR? I'd like to tweak some stuff

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
```diff
@@ -26,8 +25,24 @@ internal abstract class PlayApiService @Inject constructor(
         private val fileOps: FileSystemOperations,
 ) : BuildService<PlayApiService.Params>, OperationCompletionListener, AutoCloseable {
     val publisher by lazy {
-        credentialStream().use {
-            PlayPublisher(it, parameters.appId.get())
+        val useAppDefaultCreds = parameters.useApplicationDefaultCredentials.getOrElse(false)
+        val useExplicitCreds = (parameters.credentials.asFile.orNull != null ||
```
SUPERCILEX, line 29: This should fail even if the file doesn't exist.
```suggestion
        val useExplicitCreds = (parameters.credentials.present ||
```

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
```diff
@@ -26,8 +25,24 @@ internal abstract class PlayApiService @Inject constructor(
...
+                System.getenv(PlayPublisher.CREDENTIAL_ENV_VAR) != null)
+
+        if (useAppDefaultCreds && useExplicitCreds) {
```
SUPERCILEX, line 32: Let add another if statement that checks for specifying an impersonation without `useAppDefaultCreds`.

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
```diff
@@ -42,6 +43,30 @@ internal fun createPublisher(credentials: InputStream): AndroidPublisher {
...
+    val transport = buildTransport()
+
+    val appDefaultCreds = GoogleCredentials.getApplicationDefault()
+
+    val credential = if (impersonateServiceAccount != null) {
+        ImpersonatedCredentials.newBuilder()
+                .setSourceCredentials(appDefaultCreds)
+                .setTargetPrincipal(impersonateServiceAccount)
+                .setScopes(listOf(AndroidPublisherScopes.ANDROIDPUBLISHER))
+                .setLifetime(300)
```
SUPERCILEX, line 56: Max this out to 3600. We don't want folks with really long builds getting timed out.

SUPERCILEX (changes requested):

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
pauldthomson: It seems the tokens are generated as the task is about to make a HTTP call (rather than at the start of a build) and also automatically refreshed if "stale" (i.e. expired or within 5mins of expiry), so the thinking was that this lower value was pretty safe.. 
But an hour isn't the end of the world so happy to just max it out if preferred...

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
pauldthomson: 👍 
The other way to go would be to allow impersonation with the static creds (i.e. not ADC).. It should work just the same but perhaps it's best to not encourage this path...

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
pauldthomson: Do you mean `isPresent`??

SUPERCILEX: Yeah sorry I'm forgetting how Kotlin works :sweat_smile:

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
SUPERCILEX: Can one service account impersonate another? If yes, then sure let's allow it but maybe not document it.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
pauldthomson: I know the feeling 😛

play/android-publisher/src/main/kotlin/com/github/triplet/gradle/androidpublisher/internal/AndroidPublisher.kt
SUPERCILEX: Yeah, I don't see how a token living for an hour is going to hurt much more than 5 mins.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/PlayApiService.kt
pauldthomson: Yeah it can, but that path hasn't been implemented.. I was pondering whether we should implement it or just not allow it as suggested.. 
I think I'll go with the second one cos (as mentioned) we should encourage the ADC route

SUPERCILEX: Sure, I guess if somebody wants to use impersonation with a service account they can send a PR.

pauldthomson:
@SUPERCILEX all done

SUPERCILEX:
Thx. Can you give me PR access? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork I'd like to clean up some nits

pauldthomson:
Don't think we can as this is forked into a non-personal repo... The link above seems to allude to it being in personal repos and I don't see the option on this PR either.. Might have to just suggest the clean ups and have us do them?

SUPERCILEX:
Ah, sounds good.

SUPERCILEX (commented):
I'll do another pass tomorrow, but this what I was going to tweak for now.

pauldthomson:
> I'll do another pass tomorrow, but this what I was going to tweak for now.

Fixed those ones, let us know if there's anything else :)

PR #1164, opened 2025-07-08
Update gradle github actions

erikeelde:
Bump gradle github actions.

Documentation says we don't have to invoke wrapper validation anymore. 
https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#gradle-wrapper-validation

Long term effort was that I wanted to make ResolutionStrategy.AUTO compatible with configuration-cache.


SUPERCILEX:
Thanks!

PR #1169, opened 2025-09-26
Bump to gradle 8.9

erikeelde:
I'm getting caching issues with 8.8 and 8.10.2. 8.9 works for some peculiar reason. So I'll make a PR with it as an intermediate steppingstone. 

The tests that seem to fail are the ones that check that we're properly making use of gradles caches. Tests fail with 

```
expected: UP_TO_DATE
but was : SUCCESS
```
on those versions of gradle.


SUPERCILEX:
The up-to-date vs success tests are ensuring that we get cache hits.

Issue #1139, opened 2024-07-25
Enabling AUTO version bump causes exception in gradle


bpappin-atimi:
### Describe the bug

When you enable resolutionStrategy.set(com.github.triplet.gradle.androidpublisher.ResolutionStrategy.AUTO)
you get an error:<app>/build/intermediates/gpp/atimiQa/available-version-codes.txt (No such file or directory)

This only happens when I specifically enable that property.

#### Versions

<!--
Run this command to get all the version info you need:
```
+--- com.github.triplet.play:com.github.triplet.play.gradle.plugin:3.10.1
|    \--- com.github.triplet.gradle:play-publisher:3.10.1
|         +--- com.github.triplet.gradle:android-publisher:3.10.1
|         |    +--- com.github.triplet.gradle:common-utils:3.10.1
|         +--- com.github.triplet.gradle:common-utils:3.10.1
|         +--- com.github.triplet.gradle:common-validation:3.10.1
Gradle 8.7
|    \--- com.android.tools.build:gradle:8.5.1
|    \--- com.android.tools.build:gradle:8.5.1 (*)
```

#### Tasks executed

./gradlew publishVariantNameBundle

### Expected behavior

I expectd the version number to roll over when already taken.

### Additional context (if a crash, provide stack trace)

```
Caused by: java.io.FileNotFoundException: /Users/bpappin/Workspace/arcturus-android/app/build/intermediates/gpp/atimiQa/available-version-codes.txt (No such file or directory)
	at kotlin.io.FilesKt__FileReadWriteKt.forEachLine(FileReadWrite.kt:190)
	at kotlin.io.FilesKt__FileReadWriteKt.readLines(FileReadWrite.kt:219)
	at kotlin.io.FilesKt__FileReadWriteKt.readLines$default(FileReadWrite.kt:217)
	at org.gradle.internal.classpath.MethodHandleUtils.invokeKotlinStaticDefault(MethodHandleUtils.java:42)
	at org.gradle.internal.classpath.declarations.KotlinStdlibFileInterceptors.intercept_readLines(KotlinStdlibFileInterceptors.java:109)
	at com.github.triplet.gradle.play.PlayPublisherPlugin$applyInternal$1$10.transform(PlayPublisherPlugin.kt:441)
	at com.github.triplet.gradle.play.PlayPublisherPlugin$applyInternal$1$10.transform(PlayPublisherPlugin.kt:440)
[...]

PR #1172, opened 2025-10-15
auto versioning issue (file not found) fix

pitampoudel:
This pull request fixes this issue https://github.com/Triple-T/gradle-play-publisher/issues/1139


SUPERCILEX:
Ah, yeah but the snapshot repository changed to https://central.sonatype.com/repository/maven-snapshots/

I'll need to fix the docs.

PR #1173, opened 2025-10-30
Isolated Projects fix for Project.getCommitEditTask

mpartel-oura:
If this plugin is used in a non-toplevel Gradle project and [Isolated Projects](https://docs.gradle.org/nightly/userguide/isolated_projects.html) (#1144) is enabled, then accessing `rootProject` in `getCommitEditTask` fails.

There doesn't seem to be an obvious reason why the task would need to reside in the root project, so this PR changes it to the calling project.


mpartel-oura:
Flag added to PR. Not sure which setting should be the default.

mpartel-oura:
Thanks! Any plans for a non-snapshot release soon?

PR #1177, opened 2025-12-02
Update to Gradle 8.14.3

sproctor:
I had an issue running the tests locally. Having the fixture setup copy over my `local.properties` file fixed the issue.


sproctor:
I can't figure out the cache issue here. gradle 8.10 seems to have changed something with caching that breaks some tests.

PR #1192, opened 2026-06-02
Add retainInProgressRollout option to avoid halting a rollout when promoting

### LINKED ISSUE #1191: Add an opt-in option to retain an in-progress staged rollout when promoting
### Problem description

When promoting a release to a track that already has an **in-progress staged rollout** of a previous version, GPP **halts** that rollout.

`DefaultTrackManager.promote()` reads the *source* track's releases, relabels the track to the destination, and commits via `edits.tracks.update` — a full PUT replace of the track's `releases[]`. The destination track's existing `inProgress` release isn't carried into that committed array, so the Play API halts it.

Concretely: production has version *N* rolling out at 50% (`inProgress`). We promote *N+1* to production as a `draft`. After the promote, *N*'s rollout is `halted` and must be resumed manually in the Console.

This isn't GPP-specific — the same is reported against other API wrappers (e.g. `r0adkll/upload-google-play#179`), which confirms it comes from committing a track edit that omits the in-progress release. For teams that draft/stage the next version while the previous is still rolling out, the silent halt is surprising and disruptive.

### Potential solutions/workarounds

Add an **opt-in** promote option (default = current behavior) — e.g. `retainInProgressRollout` — that, when promoting and `fromTrack != promoteTrack`, fetches the destination track and preserves its existing releases not superseded by the promotion (same status or same version code), then appends the promoted release. The Play track model allows an `inProgress` release and a higher-version-code `draft`/`completed` to coexist, so the rollout continues uninterrupted.

I've validated the underlying API behavior (a read-modify-write that includes the existing `inProgress` release) against a real track: adding a `draft` kept a staged `inProgress @ 0.5` rollout intact.

### Additional context

Root cause is in `play/android-publisher/.../internal/TrackManager.kt` `promote()`. Happy to implement behind an opt-in flag (default unchanged) with unit + integration tests per CONTRIBUTING — filing this first to confirm desirability and align on naming/shape, as the guidelines ask.

daberni:
Resolves #1191.

### Problem

Promoting to a track replaces that track's releases via `edits.tracks.update`, which drops — and therefore halts — any in-progress staged rollout of a previous version already on the target track. Teams that draft/stage the next version while the previous one is still rolling out get their rollout silently halted. (Same root cause as `r0adkll/upload-google-play#179`.)

### Solution

New **opt-in** `retainInProgressRollout` option (DSL property + `--retain-in-progress-rollout` CLI flag, default `false`, so behavior is unchanged unless enabled). When promoting and `promoteTrack != fromTrack`, the promote track's existing releases that aren't superseded by the promoted one (same status or version code) are carried over, so an ongoing rollout keeps running.

### Changes

- Core logic in `DefaultTrackManager.promote()`.
- Threaded through `PromoteConfig` → `EditManager`/`DefaultEditManager` → `PromoteRelease` worker → extension / CLI option / `PlayExtensionConfig`.
- Added unit tests (`DefaultTrackManagerTest`) and a CLI integration assertion (`PromoteReleaseIntegrationTest`).
- README section "Retaining an in-progress rollout while promoting".

`./gradlew :play:android-publisher:test :play:play-publisher:test` passes locally.


SUPERCILEX:
Rationale for myself to understand why you wouldn't want this to always be the behavior:
* Right now if a new version is uploaded, an in-progress rollout is cancelled. If you were at X%, those X keep the version but nobody else gets it.
* With the flag, you can push the X up to 100% to finish the rollout while the new version is at Y% on a new random sample of users.
* This could plausibly be the default option, but it can't be the only option because you might think the rollout is bad and really do want to cancel the existing version.

SUPERCILEX:
And this entire feature is necessary because the current promotion logic simply takes the from track and renames it to the promote track with no logic to merge releases.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/CliOptions.kt
```diff
@@ -84,6 +84,14 @@ internal interface UpdatableTrackExtensionOptions : TrackExtensionOptions {
     )
     var promoteTrackOption: String
 
+    @get:Internal
+    @set:Option(
+            option = "retain-in-progress-rollout",
+            description = "Retain an existing in-progress staged rollout on the promote track " +
```
SUPERCILEX: This is not what the code does. It retains any non-conflicting rollouts. Not sure if this matters in any way, but I'm pretty sure it'll keep a [halted](https://developers.google.com/android-publisher/api-ref/rest/v3/edits.tracks#Status) release too for example.

SUPERCILEX (changes requested):
@daberni my only complaint is that the code carries over any non-conflicting releases rather than just inProgress ones. A test should be added for that. Take your second test and change `releaseStatus = ReleaseStatus.DRAFT,` to `releaseStatus = null,` and you should see a draft + completed with the current code. I would guess that will be rejected by the API because the completed release has a higher version code than the draft which makes no sense.

So probs just change the logic in TrackManager to explicitly allow merging with inProgress releases and we should be good.

SUPERCILEX:
Oh also I made some minor nit changes to remove the default from the args because otherwise you end up specifying the same default in a bunch of places. Should hopefully not be too much work to make compile.

play/plugin/src/main/kotlin/com/github/triplet/gradle/play/tasks/internal/CliOptions.kt
daberni: > Not sure if this matters in any way, but I'm pretty sure it'll keep a [halted](https://developers.google.com/android-publisher/api-ref/rest/v3/edits.tracks#Status) release too for example.

Thats a good point, I didn't think about that, but agree, `halted` very likely would be preserved too 👍

daberni: I now made this explicit by including all `release.isRollout()` and renaming the flag to fit the actual behaviour

SUPERCILEX:
Cool, this will be available as a snapshot build in a bit.

